PR #1, opened 2020-06-05
Added basic Donut plot

bobular:
More fine tuning of the plot needs to be done - e.g. test cases with very many categories, decimal place handling for the values and the central sum etc.

Big question remains over the usefulness of multiple data series.  I don't recommend merging until this is resolved...  I added a "Double" story to demonstrate this being called with two data series.  Should this plot type simply be configured (somehow) to refuse multiple datasets, or is there a use case?

Also, would this be used as a map marker (with the legend configured to be not shown) or would this somehow be subclassed, or would a more Leaflet-friendly version be coded from scratch?  (I, Bob, need to investigate this.)

### DIFF
--- a/src/plots/Donut.stories.tsx
+++ b/src/plots/Donut.stories.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import { action } from '@storybook/addon-actions';
+import Donut from './Donut';
+
+export default {
+  title: 'Donut',
+  component: Donut,
+};
+
+export const Basic = () => <Donut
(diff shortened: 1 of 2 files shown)


dmfalke:
Perhaps for multiple data series, we could do concentric pie charts, where the hole of the outer one is the diameter of the inner one, etc.

But I do agree that we should consider if this is actually useful or a good UI.

bobular:
Very old. We have new pie charts now. Closing.

PR #8, opened 2020-07-03
x-axis bin range investigations

bobular:
If we are using our own back ends to bin data for histograms, then we have to think about how to pass the pre-binned data through to Plotly so that we can take advantage of its user-selection tools.

User-selections would be good for creating ranges to add as application-wide filters (e.g. in a date picker or numeric variable range picker).

I soon realised that the example story "PreBinnedContinuous1" is almost completely useless.  Providing the bin ranges as string labels e.g. "0.1-0.2" means that when the user does a "box select", the x-range emitted is meaningless (well, it's 1 unit per bin, so maybe not completely).

I added the binWidth prop so that you can provide the x data as the left hand limit of each bin - see PreBinnedContinuous2.  The x range from the user selection is now meaningful.

Then I checked what happened with multiple data series in PreBinnedContinuous3 - one way to make it work is to stack the bars - so I added the mode prop for that.

Finally - I started looking at date-based data.  Bare year data needs coaxing into date format with the "xType" property passed to the layout.xaxis.type parameter.

The date handling isn't ideal...  Dates provided as "2002" are displayed as "1 Jan 2002".  If we know a priori if the data is day, month or year-resolution then I think we can set the date format appropriately.  Likewise the x-axis tics and labels could be formatted appropriately.

I added an 'action' for onSelected - so you can see what payload that has. Its range.x property is what we need, e.g.

`[ "2002-11-02", "2004-12-31" ]`  (actually the values are returned in the reverse order, oddly)

but the adapter will need to convert this into the appropriate resolution and bin boundaries, e.g. `[ "2003", "2004" ]`

Lots to discuss.

### DIFF
--- a/src/plots/Histogram.stories.tsx
+++ b/src/plots/Histogram.stories.tsx
@@ -75,7 +75,7 @@ export const PreBinnedContinuous3 = () => <Histogram
     name: 'series 2'
   }
   ]}
-  xLabel="numeric var"
+  xLabel="observation date"
   yLabel="count"
   mode="stack"
   binWidth={0.3}
--- a/src/plots/Histogram.stories.tsx
+++ b/src/plots/Histogram.stories.tsx
@@ -51,7 +51,7 @@ export const PreBinnedContinuous2 = () => <Histogram
   onPlotUpdate={action('state updated')}
   onSelected={action('selection made')}
   data={[{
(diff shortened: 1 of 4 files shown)


bobular:
Old stale branch/PR. Closing without merge.

PR #13, opened 2020-07-14
make 'type' optional in PlotlyPlot.tsx and add Bar Chart with Line Plot

linxu123-git:


jbrestel:
only small comment is that one of these is called "ScatterAndLinePlot" and the other is "BarCharWithLinePlot".   i would use "and" instead of "with".  Also, do we call it "BarChart" in other places or "BarPlot"?

linxu123-git:
searching around and think there is no differences between bar chart and bar plot, but people use bar plot more often.

src/plots/HorizontalBarChartWithLinePlot.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+   annotations: []
+    
+  };
+
+
+
+  for ( var i = 0 ; i < data.barX.length ; i++ ) {
+
+    var result : Partial<Annotations> = {
```
dmfalke, line 69: **nitpick:** please use `const` instead of `var`

src/plots/HorizontalBarChartWithLinePlot.tsx
```diff
@@ -0,0 +1,107 @@
...
+        family: 'Arial',
+        size: 12,
+        color: 'rgb(128, 0, 128)'
+      },
+       showarrow: false
+    };
+    layout.annotations!.push(result, result2) ;
+  }
+
+ var trace1={x:data.barX, y:data.barY, xaxis: 'x',yaxis: 'y', type:'bar', orientation: 'h'}
```
dmfalke, line 98: **nitpick**: please format this object (and others) following standard conventions.

Example:
```suggestion
 var trace1 = {
   x: data.barX,
   y: data.barY,
   xaxis: 'x',
   yaxis: 'y',
   type: 'bar',
   orientation: 'h'
 }

```

PR #20, opened 2020-11-04
Pie Plot

chowington:


dmfalke (changes requested):
This looks great! My only request is to move `src/plots/DefaultColorGen.tsx` to something like `src/utils/defaultColorGen.ts` (no need for the `.tsx` extension).

src/plots/DefaultColorGen.tsx
```diff
@@ -0,0 +1,17 @@
+import { defaults } from "plotly.js/src/components/color/attributes";
+
+/** A generator function for the default plotly.js colors.
+ *
+ *  Usage:
+ *    const defaultColorIter = DefaultColorGen();  // Do once
+ *    nextColor = defaultColorIter.next().value;   // Repeat
+ */
+export default function* DefaultColorGen() {
```
dmfalke, line 9: Convention is to use `camelCase` for generator functions, so maybe rename this `defaultColorGen`.

dmfalke (changes requested):
Forgot to include an inline comment.

dmfalke:
@chowington I'm going to make the suggested changes above and then merge this branch. Thanks!

dmfalke (approved):

PR #26, opened 2020-11-16
Boxplot

bobular:
- Specification implemented in full.
- Spec changed wrt upper/lower Fences/Whiskers as per [comments in Google doc](https://docs.google.com/document/d/1jSbVAMMfKax71v7bDaV3tQh3s3XuUgi4r6uuIDOKmlc/edit#heading=h.v73aliadp1m8).
- DK's opacity slider included via cut'n'paste.
- Other controls handled by Storybook Controls.
- Now uses src/stories directory
- stats-lite devDependency 

Bob's commentary on remaining issues to be provided in a review comment below.


bobular (commented):
Best tested with Storybook controls on the right hand side (but of course test all screen sizes and shapes)

Known shortcomings:
- the opacity slider takes up a lot of space
- if ISO date strings are (inadvertently?) supplied as trace labels, some unexpected behaviours may occur (see comments in Boxplot.stories.tsx)
- not yet tested for width scaling (lots of traces)
- not yet tested with large rawData (or perhaps large outliers if that ever happens)
  - may want to look at making jitter parameter depend on data size
- will this plot type be used to select ranges (e.g. along the x axis)?  Currently plotly isn't providing that button in its toolbar. Will need investigating, if we need this.
- is not independent of Plotly types - depends on Datum, for example. Appreciate guidance on this.

dmfalke (approved):
This looks great. I did make an update to `package.json`, moving some dependencies to `devDependencies`.

PR #28, opened 2020-11-18
Mosaic Plot

chowington:


dmfalke (approved):
A few thoughts:

- I think we need to get some guidance on the width question. We can probably do this over slack, or at a working group meeting.
- The working group recently settled on using the terms `dependent` and `independent` for axis data. I wonder if that would apply here as well. E.g., `exposure` -> `independent`, `outcome` -> `dependent`. Another question for the working group.
- I wonder if the bar width should be included in the tooltip somehow. 

I will go ahead and merge this so others can review on the storybook site.

PR #29, opened 2020-11-24
Histogram & Shared Plot State Hook

eikonomega:
# New
* Plot.ly based histogram plot with a variety of UI options.
* A default control panel for histogram plots.
* A variety of UI widget to compose control panels from.
* A custom hook which is used to share/manage state between control panel(s) and plot(s).

# Updated
* Rolled in latest updates to Storybook.

# Future Work
* Improve the integration of Storybook controls and histogram stories.
* Add a de-bounce feature to the binWidth slider to allow for (1) immediately responsive UX for user, but (2) avoid spamming backend servers with API requests.
* Create default control panels for all other plot types.
* Create additional control panel widgets as necessary.


dmfalke:
Overall, this is a really good start. I see that you have deviated form the props type specified in https://redmine.apidb.org/issues/41311. I think we should keep to that, as much as possible. I realize that there may be some things missing. We can address those as needed.

* I see that we left out `count` from `props.data[number].series[number]`. This should definitely be added.
* The `grouped` layout doesn't make sense for histograms, so should not be an option.
* I like the various color options you added.
* Adding the `title` props was good. We might also props for axis labels: `dependentAxisLabel` and `independentAxisLabel`.

bobular (commented):
I pushed a small change to add a debugging action to 'onSelection' (see "Box Select" in the Plotly toolbar/modebar) - so we can see what kind of x-axis Plotly is using.  As you can see the 'range' (bottom right of screenshot) is 2.5 to 4.4, which is not in the same coordinate space as the bin start/ends - it would be helpful (for later work using histograms to allow users to select ranges) if the x range values were more like 6-10.  Though we may choose instead to process the selected 'points' - though this would require parsing the string labels, also not idea.

The solution may be simply casting the x values to integers or something like that.

Screenshot_2020-12-16 In Development Histogram - Two Data Series ⋅ Storybook

src/stories/Histogram.stories.tsx
```diff
@@ -0,0 +1,335 @@
...
+        onSelectedUnitChange={plotControls.setSelectedUnit}
+        barLayout={plotControls.barLayout}
+        onBarLayoutChange={plotControls.setBarLayout}
+        opacity={plotControls.opacity}
+        onOpacityChange={plotControls.setOpacity}
+        orientation={plotControls.orientation}
+        onOrientationChange={plotControls.toggleOrientation}
+        containerStyles={{ width: 500 }}
+        binWidth={plotControls.histogram.binWidth}
+        binWidthRange={plotControls.histogram.binWidthRange}
```
bobular, line 166: A massive amount of fantastic work throughout this PR Mike - the async story data, the plot controls/hook, lovely widgets etc.  I'm still getting my head around some of it!

I was wondering if this part could be made more concise with something a bit like `{...plotControls}` but I know it's probably not that simple.

eikonomega: My friend, your wish is granted. :sparkles:

bobular:
I accidentally promoted this from draft to a real PR, but that's probably not a bad thing.

I have been reviewing the code, in the knowledge that there are things still to be committed/pushed.

My first conclusion/opinion is that we should **not** address the "axis coordinate space" w.r.t. Plotly box-select.  At the moment the x-range (or y-range if in vertical mode) emitted in the Action panel is in "bin space", which means the first bin starts at x=0, second bin at x=1 etc.  I think our client code should convert this to "data variable space" (e.g. some numbers in the thousands, in story "Single with Controls"). The client knows the bin starts/ends so I think this should work.  If we instead tried to get Plotly to plot in "data variable space" (e.g. with box select producing numbers in the thousands in that example) then we may have to manually specify bar width... That will be a pain when dealing with monthly date-based data.  We don't really want February bars to be narrower than the others.

bobular:
A possible issue is that the PlotControls might not work with date-based data.  

This, in usePlotControls.tsx (does this have any JSX in it?) could fail?  Apologies if that's already on the TO DO list.
```
          if (typeof bin.binStart === 'string') {
            console.error('String Bin Start Values are not yet supported.');
```

bobular:
Final comment from me...!

@eikonomega, do you have a plan for adding controls to another plot type, or would you prefer to leave that as an exercise for another developer after this PR is merged?  I'm in favour of either.

eikonomega:
I'm planning on adding default controls for *all* plot types.
At least, that's what I'm hoping to work on as soon as this is merged!

----------------------------------------------------------------
*Michael Dunn (a.k.a. Big Fat Panda)*
*Ave Crux, Spes Unica*

*Research Programmer*

*Center for Research Computing*

Email: bigfatpanda@nd.edu
Twitter: @eikonomega



Returning hate for hate multiplies hate, adding deeper darkness to a night
already devoid of stars. Darkness cannot drive out darkness; only light can
do that. Hate cannot drive out hate; only love can do that. - MLK

dmfalke:
Should this be considered merge-ready? If so, I can do a final review and
merge of all looks good. Or Bob can, if he would like 🙂. Let me know!

bobular:
> Should this be considered merge-ready? If so, I can do a final review and merge of all looks good. Or Bob can, if he would like slightly_smiling_face. Let me know! 

Not quite. Mike's got a few more things to do, and yes, I'll look after the merge :smiley:

bobular:
Just the streamlining of the props passing in the stories file still remaining to commit/push, right @eikonomega ?

bobular:
I just pushed some changes to src/stories/plotControls/HistogramControls.stories.tsx

My main aim was to understand the usePlotControls better by trying to fix the TS errors that were causing a build check fail.  I think I succeeded there (TS errors are gone) but I am not sure about the following:

1. Setting the initial/default binWidth in the story code
1. Was this the correct way to allow availableUnits in the story.args?
```
-export const AdditionalOptions: Story<usePlotControlsParams<HistogramData>> = (
+export const AdditionalOptions: Story<usePlotControlsParams<HistogramData>
+                     & {
+                       availableUnits: string[],
+                       selectedUnit: string,
+                     }> = (
```
3. the units widget doesn't actually work...

bobular (approved):
Looks great to merge!

PR #30, opened 2020-12-03
removed Plotly opacity slider and moved it to storybook controls

bobular:
Moving controls outside of the plot component means that the outside has to decide whether or not the control is appropriate to show.

I have put this logic in the storyArgTypes function.

I also now doubt calling many of the props defaultXXXX - that naming assumed that these were the defaults that the user could alter with a plot-internal control widget.

### DIFF
--- a/src/plots/Boxplot.tsx
+++ b/src/plots/Boxplot.tsx
@@ -18,62 +18,62 @@ export interface Props {
   }[];
   xAxisLabel?: string;
   yAxisLabel?: string;
-  defaultYAxisRange? : [Datum, Datum];
-  defaultOrientation?: 'vertical' | 'horizontal';
-  defaultShowRawData?: boolean;
-  defaultShowMean?: boolean;
-  defaultOpacity?: number;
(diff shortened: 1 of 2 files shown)


dmfalke:
@bobular I agree, defaultXXX doesn't make sense. I think you should remove the prefix since these values are set externally to this component.

dmfalke (approved):
This looks good to me. I do wonder if we should be explicit that opacity is for outliers only, but that can be addressed in a follow-up if we decide.

bobular:
Yes, probably, though it would be for all point data (outliers or rawData).

Thanks for the review

(I'm also testing replying by email)

dmfalke:
Ah, right. Maybe we can call it `markerOpacity`?

bobular:
I agree.
Was going to do a cheeky commit to master but it's protected :-(
Now I have to wind back my local commit somehow.

dmfalke:
You can make a new branch based on your local copy of master, with that
commit. Then push and make a PR. When its gets merged, you should be good
to go.

bobular:
Thanks - that was much simpler than the hoops I was starting to jump
through.

All sorted and merged now.

PR #31, opened 2020-12-07
cheeky commit to master.  renaming opacity prop to markerOpacity afte…

bobular:
…r discussion with DaveF


bobular:
Can't push changes to master, but I had changes commited to master locally.
did 
```git checkout -b new-branch```
and was able to push that to master and create this PR

PR #34, opened 2020-12-15
Add width, height, showLegend, showModebar props to plots

chowington:


bobular (approved):
Looks good and will merge - see minor concerns in Slack about reverting to default width after setting the width prop to undefined.

dmfalke:
This looks pretty good. There is one issue that should be addressed: the `layout` object is being mutated when the additional props are added. This should be changed to create a new object.

```
layout={Object.assign(layout, { ... })}
```

Should be

```
layout={{...layout, ...}}
```

PR #40, opened 2021-01-14
Add assets for map stories

dmfalke:
When migrating branches from mapveu to this repo, some static assets were missed. This PR adds them.


moontrip:
This would resolve most of issues that mapveu in this repo currently experiences, e.g., no legend, etc. I will update a small fix for margin issue after this PR is approved

moontrip:
Since this is the one I need to add a fix for margin issue, I have added the fix in this PR too.

dmfalke:
Thanks, @moontrip! I will go ahead and merge this.

PR #39, opened 2021-01-13
Entity Diagram

bfox6:
- Mini and Expanded Entity Diagrams using Visx
- Buttons allow for switching of orientation and size
- Animation using react-transition-group and css

PR #41, opened 2021-01-14
Multiple worlds solution

bobular:
Seamless east-west scrolling, gridlines are shown both sides of the 180° line in the Pacific, markers are automatically moved over that line when needed.

From now on, story or client code needs to be able to handle viewports where west_longitude > east_longitude (this happens in the Pacific).  I'll leave a comment in the PR next to an example of this.

worldCopyJump is turned off.  Users can hypothetically pan infinitely to the left or right until the longitude overflows, though I don't think anyone could in practice.


src/map/AnimatedMarkers.stories.tsx
```diff
@@ -60,20 +61,32 @@ const getMarkerElements = ({ bounds, zoomLevel }: BoundsViewport, numMarkers : n
...
+    const west = bounds.southWest.lng;
+    const east = bounds.northEast.lng;
+    const lambda = 1e-08; // accommodate tiny rounding errors
+    
     // is it within the viewport bounds?
-    if (lat > bounds.southWest.lat &&
-	lat < bounds.northEast.lat &&
-	long > bounds.southWest.lng &&
-	long < bounds.northEast.lng) {
+    if (lat > south &&
```
bobular, line 72: Here's one example of handling east < west and east > west.

moontrip:
@bobular This looks good to me as it behaves like worldCopyJump option with much smoother way 👍 
Please note that I noticed a couple of typescript errors that cause the failure of CI/build check thus I have fixed them. So now all checks have passed without error.

Just another note: I don't know why most of all past commits by others are included in this branch as the dk-bm-worldmap branch itself (at mapveu repo) would not be that far away from its master - see a similar branch like entity-diagram PR (https://github.com/VEuPathDB/web-components/pull/39) did not bring massive commits, though it seems to have a typescript error. Perhaps we should ask @dmfalke to take look at this, just in case?

bobular:
Thanks DK!
I did fix quite a few TS issues relating to MarkerProps -> BoundsDriftMarkerProps  but obviously missed a few...

bobular:
Regarding the large number of commits - I'm not sure why that is, but scrolling through the "files changed" diff here, all those changes seem only related to our multiple worlds work.  So I think it will be OK.

bobular:
There seems to be a bug still.

On opening this story http://localhost:9009/?path=/story/map-animated-markers--geohash-ids and zooming out, the grid gets stuck and this error is in the console.  It is most likely something to do with the rectangle polygons passed to shape2geohash - but I have double checked and all the first and last points are the same!

```
Uncaught Error: First and last Position are not equivalent.
    polygon index.js:218
    processRowSegment index.js:237

```

dmfalke:
I'm not sure why all of those commits are included here. The same is true of other branches from the `src/map` directory. The `entity-diagram` branch has changes in a separate directory, which is why that one has fewer commits. I suspect it's related to how the commits were merged into this repo. I don't think it will be an issue, other than having a messy commit history.

dmfalke:
FWIW, I just did a test merge on my laptop, and looks good. So, I'm not concerned about the extra commits listed here.

bobular:
> There seems to be a bug still.
> 
> On opening this story http://localhost:9009/?path=/story/map-animated-markers--geohash-ids and zooming out, the grid gets stuck and this error is in the console. It is most likely something to do with the rectangle polygons passed to shape2geohash - but I have double checked and all the first and last points are the same!
> 
> ```
> Uncaught Error: First and last Position are not equivalent.
>     polygon index.js:218
>     processRowSegment index.js:237
> ```

Now I can't reproduce it any more... :-(

moontrip:
> There seems to be a bug still.
> 
> On opening this story http://localhost:9009/?path=/story/map-animated-markers--geohash-ids and zooming out, the grid gets stuck and this error is in the console. It is most likely something to do with the rectangle polygons passed to shape2geohash - but I have double checked and all the first and last points are the same!
> 
> ```
> Uncaught Error: First and last Position are not equivalent.
>     polygon index.js:218
>     processRowSegment index.js:237
> ```

Actually I have also tested the story yesterday and at one point it might have shown an error but I could not reproduce it either

moontrip (approved):
this works good

PR #43, opened 2021-01-15
[DK] show & hide legend

moontrip:
@bobular Here is an implementation to show and hide legend via a button click: this applied to both Donut and Chart markers. I placed the action button at the bottom of legend as it may be more intuitive than other location.

I have tested several other cases including animation effect (e.g., slide in/out in the right side), but it was not successful - I think that it is because the legend itself is not made with leaflet-control, which occupies area beyond map when using animation effect (hiding)

You can test this at pre-existing stories involving Chart and/or Donut Markers.


src/map/MapVEuLegendSampleList.tsx
```diff
@@ -44,7 +44,8 @@ export interface LegendProps {
 
 //DKDK make legend at the map without using L.Control: perhaps send props to make circle or square?
 // For now, just use different component for square
-const MapVEuLegendSampleList = (props: LegendProps) => {
+//DKDK show/hide legend
+const MapVEuLegendSampleListHide = (props: LegendProps) => {
```
bobular, line 48: Please could the component naming be simpler?

MapVEuLegend or just Legend would work fine.
(and rename this file too - I don't know what "SampleList" means or why it is needed)

bobular: Given that the subcomponents start with "Legend" then I suggest we call this component "Legend"

src/map/MapVEuLegendSampleList.tsx
```diff
@@ -53,46 +54,70 @@ const MapVEuLegendSampleList = (props: LegendProps) => {
...
-        />
+    //DKDK add wrapper divs for showing and hiding legend
+    <>
+      <div ref={legendWrapper} className="legend-wrapper" >
+        <div className="legend-nav">
+        {/* DKDK add below divs for benefeting from pre-existing CSS (vb-popbio-maps.css)         */}
+          <div className="info legend">
+            <div className={legendIconClass}>
+              {/* DKDK add react-bootstrap dropdown and dynamically generate menu items */}
+              <LegendListDropDown
```
bobular, line 78: Could be LegendDropDown

src/map/MapVEuLegendSampleList.tsx
```diff
@@ -53,46 +54,70 @@ const MapVEuLegendSampleList = (props: LegendProps) => {
...
+                legendType={props.legendType}
+                dropdownTitle={props.dropdownTitle}
+                dropdownHref={props.dropdownHref}
+                dropdownItemText={props.dropdownItemText}
+              />
+              {/* DKDK legend list  */}
+              <LegendListGeneral
```
bobular, line 85: Could be LegendMain or Body or General

src/map/MapVEuLegendSampleList.tsx
```diff
@@ -53,46 +54,70 @@ const MapVEuLegendSampleList = (props: LegendProps) => {
...
+                // legendType={legendTypeValue}
+                data={props.data}
+                // divElement={div}
+                //DKDK add legendType props for handling icons
+                legendType={props.legendType}
+                //DKDK used for legend info text, e.g., Collections
+                legendInfoNumberText={props.legendInfoNumberText}
+              />
+              {/* DKDK add radio button component here */}
+              <LegendListRadioButton
```
bobular, line 95: Could be... (you get the idea!)

src/map/MapVEuLegendSampleList.tsx
```diff
@@ -53,46 +54,70 @@ const MapVEuLegendSampleList = (props: LegendProps) => {
...
+                //DKDK used for legend info texts, e.g., Collection Date, Collections
+                dropdownTitle={props.dropdownTitle}
+                legendInfoNumberText={props.legendInfoNumberText}
+              />
+            </div>
+          </div>
+        </div>
+      </div>
+      {/* DKDK button to control show/hide legend */}
+      <div className="legend-display-button active">
```
bobular, line 114: The button works great!

Visually however - on Firefox at least, the button goes transparent on mouse-over, which looks a bit weird.

bobular (changes requested):
Please could you also remove excess commented code from the story files, e.g.

in DonutMarkers.stories.tsx (and also ChartMarkers)
```
        // //DKDK send x-/y-axes lables here
        // variableLabel={variableLabel}    //DKDK: x-axis label
        // quantityLabel={quantityLabel}    //DKDK: y-axis label
        // tickLabelsVisible={knob_legendTickLabelsVisible}
        // //DKDK legend radio button props
        // onChange={legendRadioChange}
        // selectedOption={legendRadioValue}
```

moontrip:
@bobular Thank you for your review and test. I agree that the component names are quite long, so will change them following your suggestions, like, Legend, LegendMain, LegendDropDown, and LegendRadioButton.

I didn't check the button at FireFox so great to know about the issue - will check it out too

PR #44, opened 2021-01-26
Fix popups on edges

chowington:
Also fix showModebar prop


bobular:
Mergers gonna merge...

PR #46, opened 2021-01-28
Async donut stories for two-stage loading

bobular:
This is just to get the discussion started.

In this PR:

1. moved DonutMarkers.stories.tsx to /src/stories (I think maybe it's best to keep all stories together because there are stories that concern both plot and map components)
2. made two async data-getters in src/api/getMarkersFromFixtureData.tsx - one gets fully-fledged DonutMarkers, the other gets 'basic' DonutMarkers (with no second level aggregate data for the donut chart)
3. made DonutMarkers story use new Controls and the async services

Next comment will list some questions to address.


bobular:
Questions

1. To avoid reengineering the marker components, the first stage data getter returns markers with `data: [ { label: "unknown", count: sum, color: "white" } ]` so that the existing DonutMarker component draws the simple circle with a number in it.  Do we actually want a separate BasicCircleMarker component that has a different data prop, which is essentially just `{ count: number }` and this might just draw a white circle.  (And something similar for ChartMarkers.)  I'm not sure. It's just more code to maintain. It's easy enough for the data service adapter to fake the "white donut data".
2. The two-stage async request does a crude sleep(2000) between the basic and full data requests.  If you zoom in/out before the second stage has completed, the marker updates get a little confusing.  If it's easy to do, could someone explain how to cancel the unfinished calls to handleViewportChanged() when a new request is made?  Is this the same as debouncing?
3. We need to think a bit about what the Legend should show while we are waiting for the full Marker data (the legendData can only be calculated from a full marker data response.  Should it blank out and show a "loading..." indicator, or should it just show the old legend data and update when the new data comes in (as it does at the moment).

bobular:
To dos:

1. Similar implementation of the date variable ChartMarkers story.
2. Cut down the number of map stories (many are redundant once merged into master, e.g. mouse modes)
3. Convert all stories to Storybook Controls.
4.

PR #55, opened 2021-02-18
Poss solution to disable popups while dragging

bobular:


bobular:
Noticed a new behaviour (FF Linux):

A single click on a marker that has a "pop-down" popup showing temporarily makes it rotate to the "pop-up" position.

bobular:
Another bug in this branch:

In "Donut Markers->Two Requests" story - the popups stop working after the first drag.

The First Request story isn't broken.

I haven't looked at the code for explanations yet.

chowington:
I added a fix to the clicking bug. The two-requests bug still exists, but we'll take a look at it later if it ends up being necessary. Otherwise this is a nice, clean solution. Thanks Bob!

PR #56, opened 2021-02-24
Integrating PiePlot and usePlotControls

eikonomega:
# What's New

### Histogram
* Add prop to control layout of legend.
* Add prop to display/hide plotting library tools.

### Pie Plots
* Make compatible with usePlotControls.
* Improve type definitions.
* Add additional props to provide greater control over legend options and chart layout.
* Improve documentation.
* Update Storybook assets to use "controls".

## Control Panels / usePlotControls
- Integrate ability to toggle visibility of graphing library tools. For example, the Plot.ly "mode bar". This should be fairly easy to expand to other libraries that come into use. For example, Visx.
- Create default control panel for PiePlot.

### Switch
- Create Storybook assets.
- Add the ability to have a custom label.

### Misc
- General improvement to Storybook assets.

## General
- Continue to create type definitions for common use between components.
- Improve organization of Storybook assets.
- Create a utility function that creates Plot.ly legend specifications. Hopefully, this will reduce the mental trauma of dealing with them directly for our fellow developers.

# Fixes
- Address an incorrect path in `tsconfig.json`

# Dependency Updates
- Upgrade Storybook from 6.1.16 -> 6.1.20


bobular (commented):
Looks really great so I'm sorry to kick off with a report of unexpected/undesirable behaviour...

I'm interested in how the components behave when using the height/width of their enclosing element, so I tried the following in both Histogram and Pie with PlotControls:

# clear the width and height in the Storybook Controls (I have them on the right hand side)
# play around with the storybook panel width to see that it resizes nicely
# using the PlotControls, turn off 'Plotly.Controls' 

Outcome:
The placement of the PlotControls goes screwy for me in FF.
If I resize the storybook panels, the placement is correct again.

In Chrome, unsetting the height and width causes different problems for me.  The plot has basically zero height. Given that width and height are required props, I'm not surprised this is causing issues.  I'm going to start a more general discussion in the Slack channel.

One option might be to type height and width as CSS Property instead of num pixels?  I'll have a play.

bobular:
I had the quick play...

```
import { CSSProperties } from 'react';

  export type HistogramProps = {
  /** Data for the plot. */
  data: HistogramData;
  /** The width of the plot as a valid CSS string. */
  width: CSSProperties['width'];
  /** The height of the plot as a valid CSS string. */
  height: CSSProperties['height'];
```

Then tried "1000px" and "400px" in the Histogram stories and it looks the same (no surprise there).

However, "100vw" and "100vh" create a plot that takes up the full browser panel, then the PlotControls are squished up below and don't use the full width.

The screenshot below is using "100vw" and "40vh" - the controls are also squished

image


I'll start that Slack discussion

bobular:
I know there are issues with the interaction between PlotControls and Storybook Controls - and **I don't think we should even try to address them** (because users will never see this).

However, it's nice to have a story without the PlotControls (like Pie&Donut does) so we can see that the regular props work too (via Storybook Controls).  Please could you add one for Histogram?

Sorry these comments are not very high quality!

dmfalke:
While I agree with @bobular that we should address CSS dimensions, I think we should consider doing that as a follow-up. There is a lot in the PR, and I'd be hesitant to add another layer of changes. Thoughts?

dmfalke:
Unrelated to CSS, I'd like to make sure that we are satisfying the requirements of the working group, re props changes for plots. I'm pretty sure we're all good in that department, but I want to double check.

bobular:
I've had a brief look at the code and I say it's time to merge (after @dmfalke has checked the WG requirements - or feel free to delegate to me with brief instructions).

After merge, I would be keen (when time allows) to implement BoxplotControls.

dmfalke:
I did a quick comparison of the props for `PiePlot` and `Histogram` with what we have in our working group document. I don't see major departures, so I think this is good to go. Thanks!

PR #57, opened 2021-02-25
Histogram popups & improved popup functionality

chowington:


bobular (approved):
Hi Connor.

It's looking good. I've tested as vigorously as I could - can't find any more edge cases. Please go ahead and coordinate with @eikonomega about resolving and merging (since you've both worked on Histogram it's best if you do it).

I wonder (and this is a question for the React gurus) if CSS sources can be kept closer to the TSX files that use them?  And maybe there's fancy local scoping (like in Vue, lol) so that we could get rid of the popupClass prop?

Not something to address in this ticket or PR, but something to think about if it gets hard to maintain the centralised CSS (perhaps a risk of merge conflicts?).

chowington:
Thanks @bobular! There was one notable conflict: @eikonomega had added the `displayLibraryControls` prop which has the same functionality as the `showModebar` prop that's already included in all plots via `PlotyPlot.tsx`, and I added that prop to `Histogram` in my PR. I fixed that conflict by using `showModebar` in accordance with the other plots, but that possibly messed with @eikonomega's custom plot controls. Mike should take a look to see if these changes are acceptable and functional.

eikonomega:
Thank you @chowington for bringing this to my attention.

Regarding `showModeBar` I would recommend against using this as a prop name and sticking with `displayLibraryControls`.

The reason for this is that we are: 
1. Starting to use more than just the Plot.ly library for visualizations and `usePlotControls` needs to have a more generic interface so as to be usable with all visualization libraries. 
2. In general, I think we should avoid using Plot.ly specific terminology where possible to avoid a directly lock-in to that specific library. 

On a separate note - Connor, you might want to consider integrating the `PlotSpacingAddon` rather than adding a new `margin` prop. This wasn't available until my last PR merged, but you can see how it works on PiePlot.tsx now.

Finally (sorry, I blather), yes, there are ways to localize CSS definitions in React re: @bobular. While there are still times to use global CSS, whether in React or Vue, it is easy to localize in React if wanted.

chowington:
Understood, @eikonomega! Sorry for the confusion--I missed the latest and greatest recent changes :) I'll fix up the Histogram to be nice and pretty like the Pie Plot 👍

bobular:
Hi guys, I just realised that I did not check the [popups stop after dragging](https://github.com/VEuPathDB/web-components/pull/55#issuecomment-781424740) bug I reported in the branch off of this branch.

It's still there. :sob: 

I added some simple debugging (below, not committed) that shows that ondragend is not being called in the donut and chart stories with the two-stage loading.  These stories use a state variable for `bvp` and then a `useEffect` to watch for changes to async get the marker data and call setMarkers.  I can't see how or why that would interfere with the leaflet map's ondragend hook but there we go.  I wonder if we should upgrade to the new `react-leaflet` version that has `useMap` instead of `useLeaflet` and also use `useMapEvent` / `useMapEvents` - see https://react-leaflet.js.org/docs/api-map#usemap - something to discuss in meeting today.

```
    <Map
      // ...
      ondragstart={() => {console.log("set is dragging"); setIsDragging(true)}}
      ondragend={() => {console.log("set not dragging"); setIsDragging(false)}}
    >
```

PR #58, opened 2021-03-04
Range control widgets

bobular:
Added two new widgets:
1. NumericInput
2. RangeInput

The styling is not great (they are huge) but we'll fix that later.

`NumericInput` has two modes: controlled and uncontrolled.  The controlled version works a bit better when going outside the allowed range (minValue-maxValue).

`RangeInput` uses 2x `NumericInput` in controlled mode.  I'm not yet sure how it itself can be controlled externally (e.g. setting the numbers by dragging on a plot).  Maybe I'll have a look at that now.

### DIFF
--- a/src/components/widgets/NumericInput.tsx
+++ b/src/components/widgets/NumericInput.tsx
@@ -14,7 +14,7 @@ export type NumericInputProps = {
   /** Maximum allowed value (inclusive) */
   maxValue?: number;
   /** Function to invoke when value changes. */
-  onValueChange: (newValue: number) => void;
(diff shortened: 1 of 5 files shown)


bobular:
Implemented "controlled mode" for RangeInput, as demonstrated by the "Controlled Linked" story, which you can imagine as a histogram plot and the range control widget talking to each other.  Pretty happy with it, but appreciate feedback.

bobular:
Closing this PR because the #59 is a branch from here and has all the changes we need.

PR #59, opened 2021-03-06
Histogram range selection first working draft

bobular:
Got histogram selection talking to the NumericRangeInput via usePlotControls

Needs some serious tidying up - for example the range min and max are hardcoded, and the range controls are showing up in the other stories where they are not communicating with the plot.

Big question: the plot and the plot controls communicate via shared state handled by the usePlotControls reducer.  How will the outside world (e.g. the Story in this case) be able to get/set the `selectedRange`?

### DIFF
--- a/src/components/plotControls/HistogramControls.tsx
+++ b/src/components/plotControls/HistogramControls.tsx
@@ -3,7 +3,7 @@ import useDimensions from 'react-cool-dimensions';
 
 // Definitions
 import { LIGHT_BLUE, LIGHT_GRAY } from '../../constants/colors';
-import { ErrorManagement } from '../../types/general';
+import { ErrorManagement, NumericRange } from '../../types/general';
 import { OrientationOptions } from '../../types/plots';
 import ControlsHeader from '../typography/ControlsHeader';
 
@@ -14,6 +14,7 @@ import OpacitySlider from '../widgets/OpacitySlider';
 import OrientationToggle from '../widgets/OrientationToggle';
 import SliderWidget from '../widgets/Slider';
 import Switch from '../widgets/Switch';
+import NumericRangeInput from '../widgets/NumericRangeInput';
 
 export type HistogramControlsProps = {
   /** Label for control panel. Optional. */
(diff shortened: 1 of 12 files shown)


dmfalke:
> Big question: the plot and the plot controls communicate via shared state handled by the usePlotControls reducer. How will the outside world (e.g. the Story in this case) be able to get/set the `selectedRange`?

The `usePlotControls` hook returns an object that has the current state of the plot controls, and functions for updating values. This gives you an imperative hook to get/set particular values. Having not reviewed the code yet (sorry!), you may need to add these items to the return value of the hook.

One thing that might not work right now is updating the value if the initial state passed to `usePlotControls` changes (e.g., a different range is passed in). This will need to be addressed. I may do that in concert with the filter integration.

bobular:
Thanks Dave.

I believe I have sorted the "external communication" issue and a range of other niggles.

The code has been through quite a few transformations, so it's good you haven't done a review yet!

Here are some remaining issues/questions:

- binEnd calculation is messy (a choice between splitting the label or adding the binwidth - which may be tricky with dates) - maybe we should ask for it from the back end?
- it's also messy here because the demo data doesn't have the same bins for all series (but the EDA data does/will)
- the next big task is to make this all work for numeric or date.  This will be best using DateInput fields and  DateRange type etc.  a medium-large task but hopefully nothing too tricky.  Merge before or after this task?
- does usePlotControls somehow need to be modularised? It now has quite a lot of pie and histogram-specific code. It may become difficult to maintain.  Though I hope modularisation wouldn't add any more complexity!  It took me a few days to get my head around it fully.

dmfalke:
This all sounds great!

Re merging: I think we should wait. I can always build something using a branch of this repo. Do you anticipate the code on this branch to work while you're working on the date stuff? If not, maybe a new branch based on this one? When it's ready, we can merge that one into `master`.

Re modularizing usePlotControls... I think we should, but it's not urgent. I have ideas on how to do this. We can discuss when we get some breathing room.

I will try to look at what's in this PR soon. Maybe we should make it a draft again to prevent accidentally merging?

bobular:
Thanks Dave.

On reflection (and your comments) I've made one more change which simplifies usePlotControls (slightly).

I've made it a draft PR and will make a new branch for the date work.

bobular:
Superseded by a later PR

PR #62, opened 2021-03-12
Continuous histogram axis, date histograms, date range selection

bobular:
All pretty much working.

When selecting a date range (**Date Range Selection** story) that extends all the way to the right hand edge, it gets stuck in an infinite loop.  There seem to be some strange edge cases with dates (see the 30/Apr default range lower bound - it should be 1/May).  Will have a closer look with a fresh head.

There's also probably another bug: when you select a range in the numeric histogram it's quite snappy, but when you select a range in the date-based histogram, it flashes for a second before showing the selection highlighted.  Needs investigating, as maybe there is some inefficient code somewhere.

I would like to wire the binWidth and binWidth units controls in properly but I think that's a big job that could be done at the same time as implementing zoom (they are related... as you zoom in you need to change the binning units).


bobular:
There's also a UX issue.  After making a selection, there is no way to tweak it. You have to make a whole new selection. It would be good if the Plotly selection handles and boundaries didn't disappear.  Not sure why they do, yet!

bobular:
Also had some overnight thoughts about histogram controls.

I think it needs a fresh think/overhaul.

For numeric variable histograms, I can't think of a need for `selectedUnit`.  If showing a histogram of person height handled by the database/back end in cm, do we really need to provide binning controls in m as well?  Surely not urgently.

More generally, for other plot types, e.g. bar chart, I can imagine a plot of height vs female/male:  the user might want to change the y axis scale from cm to m but that is also surely a phase 2 or 3 thing and nothing to do with histograms. 

So I suggest we focus instead on date variables (and histograms)...

Here it's all about bin widths.  Even before we implement zooming, the user may genuinely want to bin at, say, a monthly or yearly level. We can't use fixed millisecond-based widths because the bins are different widths for months and years (due to leap years).

So for date-variables, histogram bin width is always a `TimeDelta` amount (number plus date unit), and we need a `TimeDeltaSlider` component (analogous to the DateInput (vs NumberInput) component).  So I suggest we remove `onSelectedUnitChange` from the equation for now, and move any "unit stuff" inside the histogram hierarchy of usePlotControls.

There would still need to be the concept of `availableUnits` for the unit selector in the `TimeDeltaSlider` widget, and the available units would need to be derived from the range of data (min binStart to max binEnd) and perhaps the number of pixels available in the independent axis.

bobular:
One more thing... this branch expects bin.binStart and bin.binEnd from the back end.  Danielle has agreed to provide this but I don't know the timeline.

Without binEnd there were just too many hoops to jump through, and in one case, having to guestimate the right-most binEnd.

dmfalke:
I mostly agree with this. Units may not be the right term. Maybe time
resolution? Anyway, keeping this separate from units would be helpful.

dmfalke (approved):

PR #65, opened 2021-03-15
Initial selected range

dmfalke:


src/hooks/usePlotControls.tsx
```diff
@@ -321,16 +323,20 @@ export default function usePlotControls<DataShape extends UnionOfPlotDataTypes>(
...
-        .sort(
-          (a: NumberOrDate, b: NumberOrDate) => a.valueOf() - b.valueOf()
-        )[0];
-      const max: NumberOrDate = params.data.series
-        .map((series) => series.bins[series.bins.length - 1].binEnd)
-        .sort(
-          (a: NumberOrDate, b: NumberOrDate) => b.valueOf() - a.valueOf()
-        )[0];
+      const min: NumberOrDate =
+        params.histogram?.selectedRange?.min ??
```
bobular, line 327: We're calculating the **bounds** for the selected range here (i.e. the min value for the left hand text box, and the max value for the right hand text box).  Therefore we don't want to use `params.histogram.selectedRange` here.  The bin and max should be calculated from the data every time.

I noticed this causing unwanted behaviour in the front end: after selecting a range, you can't then select a wider range.

dmfalke: I disagree! This supports the case where we render a histogram that already has a filter associated with that variable. For instance, coming back to an existing session. I admit this was a little sloppy and rushed. But in general, we should support the consumer passing in any values to override defaults.

dmfalke: Did I misinterpret how this was being used? I think I did! :blush:

PR #86, opened 2021-03-30
Storybook control-ification of EntityDiagram

bobular:
Note: PR merges to `entity-diagram`

Just wanted to show how to make stories "Controls" compatible.

Now it's easy to adjust the size and the shading numbers.

I didn't use the Template and Template.bind() cloning approach, but that would be easy to add if more stories were needed (though with Controls, they often aren't).  See Boxplot.stories.tsx for examples using template cloning.

I removed the "separate controls" story and imports because I figured the merge would be easier if Connor has already started that bit.

### DIFF
--- a/src/EntityDiagram/EntityDiagram.stories.tsx
+++ b/src/EntityDiagram/EntityDiagram.stories.tsx
@@ -1,14 +1,16 @@
 import React, { useState } from 'react';
-import MiniDiagram from './MiniDiagram';
-import ExpandedDiagram from './ExpandedDiagram';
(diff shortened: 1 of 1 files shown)


chowington:
Thanks Bob! Actually I also storybookified the ED yesterday after pushing the previous PR. 😅 I'll compare your and my version and merge if needed

bobular:
Cool - delete this as needed.  I spotted one item in shadingData was missing: GEMS_Sample I think, not that that mattered!

Issue #70, opened 2021-03-15
Add zooming controls for histogram x-axis


bobular:
There's currently functionality for **range selection** in the histogram and histogram controls

https://veupathdb.github.io/web-components/?path=/story/plots-histogram--range-selection

To add zooming in the x-axis, we could

- add a new control that selects between "select" (default) and "zoom" behaviour.
- when in zoom mode, a "selection" will trigger a new back end request (see API definition for `{date,numeric}-histogram-{bin-width,num-bins}` here: https://veupathdb.github.io/EdaDataService/api.html#resources:/analyses/numeric-histogram-bin-width:post ) with the config.viewportMin and config.viewportMax set.

see also #68 for resetting all the ranges and zooming out.

dmfalke:
I don't agree with toggling the functionality of a range selector as described in the description. I think these should be distinct widgets. @moontrip and I have discussed this and he is incorporating this into #87.

bobular:
Does this mean having just text input-based zooming? (not mouse-based)

dmfalke:
Yes, that was my understanding of what we agreed upon. I could be remembering incorrectly. I think we will want the text inputs regardless, so we can always add a toggle to change the behavior of the mouse gesture in the future.

fwiw, I didn't fully understand that this issue was about the mouse gesture, which may have added to some confusion.

bobular:
Old ticket - but perhaps a "zoom to selected range" button (in the EDA's HistogramFilter) might be a good compromise?

In EDA's HistogramVisualization we could quite easily hook up the currently inoperative range-selection to set the x-axis range, if we want to.

Issue #68, opened 2021-03-15
Histogram: add a control to reset the selected range


bobular:
This could be a general "reset to defaults" button.

See also the zooming ticket to be added soon

Issue #66, opened 2021-03-15
Histogram: allow adjustment of selected range


bobular:
In [this story](https://veupathdb.github.io/web-components/?path=/story/plots-histogram--range-selection) the selection "handles" disappear after selection and the selection is transformed by the Histogram component into a more opaque set of bars.

There is no way to adjust the range without drawing a whole new range, which isn't great UX.

bobular:
This has been interpreted as "text entry boxes don't allow editing or input of selected range"

This is true and is a more important bug to fix than the original bug which refered to graphical range selection (and adjustment of previously selected range wasn't a feature in ClinEpi 1.0)

Issue #61, opened 2021-03-11
Histogram: rename yAxisRange prop to dependentAxisRange and make it NumberRange type


bobular:
Has quite a few downstream consequences: e.g. `ChartMarkers`

Then in `ChartMarkers.stories.tsx` it's call yAxisRangeValue which could be renamed dependentAxisMaxValue or similar - or maybe pass a min/max range object rather than just the max to keep things consistent.

bobular:
If we are going to be completely consistent about x and y being independent and dependent, there's an xAxisVariable in the data service API specs for histogram at least: https://veupathdb.github.io/EdaDataService/api.html#resources:/apps/pass/visualizations/numeric-histogram-bin-width:post 
This would need front+back end coordination if we decide to fix.
@d-callan

bobular:
Sorry I missed this on the first review.

There's still `yAxisRange : [number, number]` in Histogram.tsx - this needs renaming please, to be consistent with the work done for #87

bobular:
I had an idea about the empty subsections.

The subsection containers can be handled by a component that renders nothing when `children.length === 0` (or similar).

moontrip:
This is made at #87 PR

PR #87, opened 2021-03-30
[issues-69] add y-axis controls

moontrip:
tried to add y-axis controls. Checked that the states of log scale (on/off) and absolute/relative changed per their UI (button/buttonGroup) actions (Histogram->Y Axis Controls story). The min/range supports both number (can be seen at Required Options and Y Axis Controls stories) and date (can be found at Additional Options story). A screenshot is attached here.

Image 5


src/components/plotControls/HistogramControls.tsx
```diff
@@ -276,9 +296,66 @@ export default function HistogramControls({
...
+      <div>
+        {
+          <ControlsHeader
+            text={'<Y-Axis Controls>'}
+            styleOverrides={{ paddingTop: 30, textAlign: 'left' }}
+          />
+        }
+      </div>
+      <div style={{ display: 'flex', flexWrap: 'wrap', paddingTop: 0 }}>
+        {valueType === 'number' && toggleYLogScale && yLogScale !== undefined && (
```
bobular, line 314: log scale on the dependent axis should be OK for any valueType (number or date).  valueType is only describing the independentVariable.  Maybe it should be renamed to reflect that?

The conditional display of control widgets should only depend on
a) the true-ish state of the callback (here toggleYLogScale)
b) presence of other props required to draw the widget

See how Mike/Dave did it for the other widgets above: orientation, barLayout, selectedUnit

dmfalke:
This is looking good. I think the layout could use some work. I don't have any particular suggestions at the moment. A quick thought is that we should have a section for **X-AXIS CONTROLS**, in addition to **Y-AXIS CONTROLS**.

src/hooks/usePlotControls.tsx
```diff
@@ -454,6 +500,18 @@ export default function usePlotControls<DataShape extends UnionOfPlotDataTypes>(
     }
   };
 
+  // Toggle y-axis logScale
+  const toggleYLogScale = () => dispatch({ type: 'histogram/toggleYLogScale' });
```
bobular, line 504: Because usePlotControls always returns a `toggleYLogScale` callback, the LogScale widget is shown in all the stories, not just the 'y-axis controls' story.

We'll discuss removing plotControls from the story altogether at this week's meeting.

bobular (commented):
Thanks for the PR!

I think more general discussion about plotControls is needed before going much further (but you could start the renaming and minor mods already).

src/components/plotControls/HistogramControls.tsx
```diff
@@ -90,6 +90,19 @@ export type HistogramControlsProps = {
   accentColor?: string;
   /** Attributes and methdods for error management. */
   errorManagement: ErrorManagement;
+  // add y-axis controls
+  /** Whether or not to show y-axis log scale. */
+  yLogScale?: boolean;
+  /** Action to take on y-axis log scale change. */
+  toggleYLogScale?: (yLogScale: boolean) => void;
```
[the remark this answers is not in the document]
moontrip: @bobular I have complied with your suggestions. Also, reset button (UI only) ([issues #68](https://github.com/VEuPathDB/web-components/issues/68)) was added as well.

src/components/plotControls/HistogramControls.tsx
moontrip: @bobular following your suggestions, the conditions are simplified without checking valueType

bobular (commented):
Looking good DK - I think maybe we should decide on the future of usePlotControls before merging (or working more) on this branch.  There's a topic on the agenda for tomorrow's meeting (which I can't attend, unfortunately).

dmfalke:
My main issue is still with the layout. I think it's difficult to parse. I think we should organize the individual controls into three main sections: **Y-Axis**, **X-Axis**, **General** (in that order). I also suggest we standardize on having labels above controls at all times. We can probably also get rid of the ****MIN**/MAX** labels for range selectors. They add an additional line and cause controls to be misaligned. Perhaps we need an option to hide the labels in the range widget.

Just for reference, this is an example of what we currently have on clinepidb.org:
image. I'm not suggesting we copy this, but it shows how we are clearly segregating x- and y-axis controls.

moontrip:
@dmfalke I have made a mockup like attached. Two questions arise: a) should "Log scale" be radio button style with texts like your screenshot?; b) should "Range" be aligned with input fields? It is doable but such a change will affect all range input forms like "Selected Range" etc.

y-axis1

dmfalke:
This looks really good. I think the switch widget is fine for log scale. And I think putting range on the same line as the inputs is a good idea, if easy. Nice work!

moontrip:
@dmfalke Thank you for your review. I tried to make the range in-line but before committing my changes, I want to listen to your opinion: please see the below screenshot and see "SELECTED RANGE" at the top also is changed. Just let me know if you prefer this than two lines Range input fields :)

y-axis4

dmfalke:
Thanks @moontrip. I think two lines is better. Let's go with that. Thanks!

moontrip:
@dmfalke Thank you for your reviews and suggestions 👍 Corresponding commit is made with the new layout. Here is a screenshot for this commit which is very similar to earlier one (added some margin/padding)

y-axis6

moontrip:
@dmfalke I realized that I did wrongfully use the reset all button to y-axis reset button. Will add like dependentAxisRangeReset button and use it for y-axis button.

dmfalke:
This is looking great! Do you plan to add the x-axis controls in this PR as well?

moontrip:
@dmfalke Thank you for your suggestion! Probably it is alright to use this ticket to add x-axis (independent axis) one at least. Will do that tomorrow morning :)

dmfalke:
> @bobular and @dmfalke, Following Dave's suggestion to add x-axis control group like I did for y-axis one above, I want to make sure of what kinds of controls should be in there. I think that Data Units (button group) , Bin Width (slider), and Selected Range (range) belong to x-axis control group, right? Initially I planned to use the term, independentAxis... for x-axis ones, but it may cause lots of additional changes to pre-existing ones (e.g., EDA Viz one, etc.), thus for now it is better to keep the current names (e.g., selectedRange, etc.) but group them under the umbrella of x-axis control box. What do you think?

I think those controls mostly sound right. I think selected range should remain outside of axis-specific controls. But the visible range should be added. And we should stick with the **x-axis** label for now.
 
> Also I was thinking of additional props to show/not to show the control group (x- or y-axis) (e.g., showDependentAxisControls), but it appears to me that each plot will probably have its own control component like HistogramControls component for histogram. Thus, such props seem not to be needed unless we will have a GeneralControls component.

Yeah, we can see how things look with other plot types.

moontrip:
@dmfalke Thank you for your information. I have update the overall layout and added a new x-axis range (named as independentAxisRange...). Here are two screenshots for number and date (with Data Units), respectively.

axis-number1

axis-date1

src/components/plotControls/HistogramControls.tsx
```diff
@@ -276,9 +317,199 @@ export default function HistogramControls({
...
+                toggleDependentAxisLogScale(event.target.checked)
+              }
+              containerStyles={{ paddingBottom: 5 }}
+            />
+          )}
+          {onDependentAxisRangeChange ? (
+            valueType !== undefined && valueType === 'date' ? (
+              <DateRangeInput
+                label="Range:"
+                // rangeBounds={selectedRangeBounds as DateRange}
```
bobular, line 373: on the dependent (y) axis, it's not appropriate to use `selectedRangeBounds` as bounds for the range controls - because `selectedRangeBounds` is in independent (x) space.

I know it's commented out, but this suggests it could be uncommented later, which would be misleading.

So I think we should remove all these commented-out  `rangeBounds={...}` props

src/components/plotControls/HistogramControls.tsx
```diff
@@ -276,9 +317,199 @@ export default function HistogramControls({
...
+            style={{
+              width: 50,
+              marginTop: '-1.8em',
+              marginLeft: '-.3em',
+              marginBottom: '.3em',
+              background: 'white',
+              textAlign: 'center',
+            }}
+          >
+            y-Axis
```
bobular, line 352: (the same applies to the x-Axis box)

I think an empty x-Axis box will be displayed if none of the contained widgets are active.  Unfortunately, I think a large conditional (toggleDependentAxisLogScale || onDependentAxisRangeChange || onDependentAxisModeChange ...) is needed around the whole div.  This won't be great for maintenance (adding a new widget requires an edit in two places) so alternative suggestions gratefully received.

src/components/plotControls/HistogramControls.tsx
```diff
@@ -180,16 +218,18 @@ export default function HistogramControls({
             selectedOption={barLayout}
             // @ts-ignore
             onOptionSelected={onBarLayoutChange}
+            containerStyles={{ paddingRight: 25 }}
           />
         )}
-        {availableUnits?.length && selectedUnit && onSelectedUnitChange ? (
+        {/* commend this out as it moves to x-axis control group */}
```
bobular, line 224: Please remove the commented code that is already added back in another place.

src/components/plotControls/HistogramControls.tsx
```diff
@@ -226,7 +266,8 @@ export default function HistogramControls({
             color={accentColor}
           />
         )}
-        {onBinWidthChange && (
+        {/* commend this out as it moves to x-axis control group */}
```
bobular, line 269: Please remove the commented code that is already added back in another place.

bobular (changes requested):
I was wondering if we can merge this without a functional story to show all the controls working, but I think we can, after addressing the small issues above.

I noticed that there's some unfinished business with https://github.com/VEuPathDB/web-components/issues/61 which I'm going to reopen.  Basically, at some point, the HistogramControls and Histogram need to share the same independentAxisRange prop, but in Histogram.tsx it is still called yAxisRange (and, more importantly, is not yet a NumberOrDateRange).

src/components/plotControls/HistogramControls.tsx
moontrip: @bobular thank you for your review and comments 👍 
I will remove them

src/components/plotControls/HistogramControls.tsx
moontrip: Yes I know the potential issue as I have made a comment earlier at https://github.com/VEuPathDB/web-components/pull/87#issuecomment-815767559, but citing in the following:

"Also I was thinking of additional props to show/not to show the control group (x- or y-axis) (e.g., showDependentAxisControls), but it appears to me that each plot will probably have its own control component like HistogramControls component for histogram. Thus, such props seem not to be needed unless we will have a GeneralControls component."

I thought that we would always show x- and y-axis controls for histogram? If there is a use case to hide those controls for histogram, then I think it is better to introduce a prop like my earlier comment above

src/components/plotControls/HistogramControls.tsx
moontrip: yep and same to the below comment :)

src/components/plotControls/HistogramControls.tsx
bobular: I had an idea about the empty subsections.

The subsection containers can be handled by a component that renders nothing when `children.length === 0` (or similar).

But you may be correct about always showing x- and y-axis controls for a histogram!

dmfalke: Don't quote me on this, but it might be possible to use a css rule like this:
```css
.AxisControlContainer:empty {
  display: none;
}
```
See https://developer.mozilla.org/en-US/docs/Web/CSS/:empty

bobular: Hi DK - did you decide on a solution for the "empty subsection" issue or are we going to skip this for now?

moontrip: Hi @bobular I have tested the CSS but it does not work properly, perhaps because the condition of "empty" is quite strict? (even no single space allowed). I think it is better to be handled at another ticket. 

Currently working on #61 within this PR

moontrip:
- merged master
- #61 concerning yAxisRange props: relevant component props are changed from yAxisRange to dependentAxisRange, including Histogram component. Some local variables using the word, yAxisRange..., are not changed

dmfalke (commented):
Looking at this picture, there are some things that can be improved:
image

1. x- and y-axis controls should be side-by-side. Maybe put them in a flex container?
2. The reset button text should not wrap
3. Generally, prefer to use `em` rather than `px` lengths.

Also, I'm not crazy about inline styles, but they are okay for now. We can address as a follow-up. I started playing with `useStyles` from the material-ui package, which adds theme-based support, which might be really useful in the future. Until we assess exactly what direction we want to go (traditional css/scss vs css-in-js), we can leave the inline styles as-is.

moontrip:
@dmfalke actually they are side by side via inline-flex: I can see both correctly at (Plot Controls) Histogram->Y-Axis controls under both Chrome and Firefox as I showed in my previous screenshots. Please note that I didn't put widths for each group (y- or x-axis box) and it depends on the width of the outer box. But the default value (or min width) of the range input field by the Material UI may be a bit larger though. So, if the outer box has smaller width, then the y- and x-axis boxes will be located at upper and lower sides, instead of side-by-side: this is the case for (Plots) Histogram->Range Selection. 

As for the reset button, I found strange that it shows correctly at Chrome but not at Firefox. Perhaps the width of the button may be a bit smaller: will check it out

dmfalke:
I see. I was looking at a histogram story. I think it was the first one.

moontrip:
@dmfalke I found a reason that the range input fields (with the text "TO") is comprised of dispay:flex thus they occupies all width depending on the control box size. A quick solution is to set width for the box, like width: 480 (30em) for both control boxes and revisit more detailed layout in the near future. Also fixed the reset button one.

For the "em", yeah actually I used px and em together for the control boxes. Since there are lots of usage for px at other places beyond x- and y-axis control boxes at <HistogramControls> component, I tried to change them to em as many as possible. 

I have pushed those changes so can you please check them out?

dmfalke (approved):
Thanks @moontrip, the reset buttons look better. The axis controls are still appearing on separate lines for me, but maybe it will be okay on the eda website and we can address this some other time.

Are there any other remaining TODOs for this PR?

moontrip:
@dmfalke Thank you for your quick review. It is quite strange that I can see them correctly at both Chrome and Firefox. We will see :)

I think no TODOs are left in this issue as I also addressed Bob's request in this PR too (https://github.com/VEuPathDB/web-components/pull/87#pullrequestreview-632255527)

dmfalke:
@moontrip are these issues related?
* https://github.com/VEuPathDB/web-components/issues/70
* https://github.com/VEuPathDB/web-components/issues/66

moontrip:
@dmfalke #70 involves mouse-based control for zooming the selected range in the plot, while #66 is text-based control of selected range only (no zooming involved). Thus, those two are technically different.

dmfalke:
Ah, ok. We can keep #70 open. #66 was reinterpreted as fixing a bug where the text box was not editable. Are you saying that wasn't addressed in this branch? I thought you mentioned you were working on it, and I assumed it was in this branch.

dmfalke:
I'm going to merge this. If #66 is still an issue, I think it will be addressed by PR #99.

dmfalke (approved):

moontrip:
@dmfalke Ah... there are so many similar numbers for issues so you might have been confused :) The one Bob raised before was #61, which I have added fixes in this PR, not #66. The latter is more related to the slider one you worked #99, which I have independently been working with my approach. Honestly, I am not quite sure if mine would be better than the solution addressed at #99 but I will make a PR today anyway - will make you as a reviewer :)

dmfalke:
Okay. I just misunderstood. When you said you were working on it, I thought you meant as a part of this PR. No worries!

PR #93, opened 2021-04-02
Fix entity diagram story

dmfalke:
Primary change is adding the `component` property to the Meta descriptor. Not sure this will work, but worth a shot.


dmfalke:
It turns out the previous PR merge deployed a fully functional site. I will still merge this for this typescript updates and script update in package.json.

PR #95, opened 2021-04-07
Small histogram range selection bug fix

bobular:
data.valueType can be undefined which wasn't being handled properly.

For the attention of @moontrip who will find this relevant to https://github.com/VEuPathDB/web-components/issues/66


moontrip:
@bobular yes I knew the bug and actually I have been using a slightly different solution at my local, by adding additional condition like `data.valueType === 'undefined'` but your approach would work too

PR #98, opened 2021-04-08
Blue rectangle-based range selection in histograms

bobular:
Good to merge.

Known limitations

- After making a selection, the user might be tempted to cancel it by making a small selection that includes zero bins. However, it's not possible to 'unset' selectedRange by passing undefined to onSelectedRangeChange (it expects a defined argument).  This can be solved more generally with reset buttons maybe.
- Range selection is still half-active in the viz panel where it's not needed. You see the graphical selection, but upon finishing the selection nothing happens (which is correct)


src/plots/Histogram.tsx
```diff
@@ -177,31 +170,153 @@ export default function Histogram({
...
+
+        // now snap to bin boundaries using same logic that Plotly uses
+        // (dragging range past middle of bin selects it)
+        const leftBin = binSummaries.find(
+          (bin) => rawRange.min < bin.binMiddle
+        );
+        const rightBin = binSummaries
+          .slice()
+          .reverse()
+          .find((bin) => rawRange.max > bin.binMiddle);
```
chowington, line 258: I'm not convinced that including bins based on whether their middle is selected is intuitive. I tried out including a bin if any of its range is selected, and that feels far better to me. Here's that version for reference if someone else wants to try it:

```suggestion
        const leftBin = binSummaries.find(
          (bin) => rawRange.min < bin.binEnd && rawRange.min > bin.binStart
        );
        const rightBin = binSummaries
          .slice()
          .reverse()
          .find((bin) => rawRange.max > bin.binStart && rawRange.max < bin.binEnd);
```

src/plots/Histogram.tsx
```diff
@@ -177,31 +170,153 @@ export default function Histogram({
...
+  const binSummaries: BinSummary[] = useMemo(() => {
+    const allBins: HistogramBin[] = data.series
+      .map((series) => [series.bins])
+      .flat(2);
+
+    const seenLabels: string[] = [];
+    const uniqueBins = allBins.filter((bin) => {
+      const seenBefore = seenLabels.indexOf(bin.binLabel);
+      seenLabels.push(bin.binLabel);
+      return seenBefore < 0;
```
chowington, line 204: I think this could be more efficient? Maybe doesn't matter though, even with lots of bins

```suggestion
      if (seenBefore < 0) {
        seenLabels.push(bin.binLabel);
        return true;
      }
      return false;
```

src/plots/Histogram.tsx
bobular: Interesting, thanks Connor.

BTW, the middle-based behaviour is what Plotly does, and what we have tried to emulate.

I tested your version, and I think it tends to include an extra bin to the left of where you start dragging (left to right in my case). But maybe I got used to the Plotly behaviour...

There's also an issue with drag-selecting to the L or R extremes of the plot.  The selection is lost.  It can be fixed with the >= and <= added as below, but only for the numeric histogram (test via /story/plots-histogram--range-selection) - the date-based one still has problems (/story/plots-histogram--date-range-selection) - but maybe that can be fixed - it could be 00:00 vs 23:59:59.999 issue.

```
        const leftBin = binSummaries.find(
          (bin) => rawRange.min < bin.binEnd && rawRange.min >= bin.binStart
        );
        const rightBin = binSummaries
          .slice()
          .reverse()
          .find((bin) => rawRange.max > bin.binStart && rawRange.max <= bin.binEnd);

```

src/plots/Histogram.tsx
bobular: Definitely more efficient - thanks!

src/plots/Histogram.tsx
```diff
@@ -177,31 +170,156 @@ export default function Histogram({
...
+      return false;
+    });
+    // sort them on binStart just in case
+    uniqueBins.sort((bina, binb) =>
+      bina.binStart > binb.binStart
+        ? 1
+        : bina.binStart === binb.binStart
+        ? 0
+        : -1
+    );
```
dmfalke, line 216: **Suggestion:** This can be done with utilities from lodash, which are typically very optimized.

```suggestion

import { sortBy, sortedUniqBy } from 'lodash';
const sortedBins = sortBy(allBins, bin => bin.binStart);
const uniqueBins = sortedUniqBy(bin => bin.binLabel);
```

dmfalke (changes requested):
I played with this, and I'm pretty happy with it. I think the visual cue as to what is being selected is hugely helpful. I did add a couple of inline suggestions.

src/plots/Histogram.tsx
chowington: Good points! Dave agrees with you also, so I'll happily consent 👍

bobular:
The requested changes are done and tested.

Upon merging in `main` some UX needs attention:
image

bobular:
All good now - please approve!

image

dmfalke (approved):

src/components/widgets/NumberAndDateInputs.tsx
```diff
@@ -92,10 +95,11 @@ function BaseInput({
...
-        setErrorState({
-          error: true,
-          helperText: `Please enter a ${valueType}.`,
-        });
+        required &&
+          setErrorState({
+            error: true,
+            helperText: `Please enter a ${valueType}.`,
+          });
         return false;
```
dmfalke, line 103: I think we should notify if `required` is `false`. Presumably, in this case, `undefined` is a valid value for the consumer.

```suggestion
        return !required;
```

dmfalke (changes requested):
Sorry, I found a small glitch that presented itself with the **Linked Pair** story. When I tried to remove the value in **A** using `BACKSPACE`, I had to hit it twice to remove the last digit. I've requested a change that will address this. Please take a look. Thanks!

Edit: I should clarify that the value was not selected when doing what was described above. These are the steps I took:
1. Click to put focus in input **A**
2. Type in the value `20`
3. Hit `BACKSPACE` twice
4. The value `2` is still in input **A**

src/components/widgets/NumberAndDateInputs.tsx
dmfalke: Wait, I just realized I was reviewing the wrong feature! This should have been added to https://github.com/VEuPathDB/web-components/pull/99. I will resolve this conversation and open a new issue. Sorry!

dmfalke (approved):

src/components/widgets/NumberAndDateInputs.tsx
dmfalke: Ah, wait again! This change was made in this PR. I will unresolve this :smile:

dmfalke (changes requested):
See https://github.com/VEuPathDB/web-components/pull/98#pullrequestreview-636238451

bobular:
No worries - and thanks for finding this edge case.  Sorry it's in a PR for another issue...

I added your change, and it works great for the non-required linked pair.
I added a story for a required linked pair.  The behaviour is passable but not ideal.
If you enter 100000000 in a box, let it update the other, then hit backspace, the last valid value (1) is notified to the consumer before the box is emptied.

dmfalke (approved):
I'm okay with this new edge case. I think it's the intended behavior from the implementation perspective. The basic contract is: if the user enters an invalid value, then we will not call the onChange handler. That's what is happening here, so I think we're good.

PR #99, opened 2021-04-09
Slider with auxiliary text input

bobular:
Just realised I spelled auxiliary wrong in the story file...

I made the draft PR to make it easier to discuss a change that @dmfalke made that seemed to have broken all number/date text input fields.  I'll make a specific comment below.

### DIFF
--- a/src/components/widgets/NumberAndDateInputs.tsx
+++ b/src/components/widgets/NumberAndDateInputs.tsx
@@ -83,7 +83,7 @@ function BaseInput({
     } else {
       setErrorState({ error: false, helperText: '' });
     }
-    return undefined;
+    return newValue;
   };
 
   useEffect(() => {
--- a/src/components/widgets/NumberAndDateInputs.tsx
+++ b/src/components/widgets/NumberAndDateInputs.tsx
@@ -96,7 +96,7 @@ function BaseInput({
           error: true,
           helperText: `Please enter a ${valueType}.`,
         });
-        return undefined;
+        return false;
       }
       if (minValue !== undefined && newValue < minValue) {
         newValue = minValue;
@@ -104,17 +104,17 @@ function BaseInput({
           error: true,
           helperText: `Sorry, value can't go below ${minValue}!`,
         });
-        return newValue;
+        return false;
       } else if (maxValue !== undefined && newValue > maxValue) {
         newValue = maxValue;
         setErrorState({
           error: true,
           helperText: `Sorry, value can't go above ${maxValue}!`,
         });
-        return newValue;
+        return false;
       } else {
         setErrorState({ error: false, helperText: '' });
-        return newValue;
+        return true;
       }
     },
     [minValue, maxValue]
@@ -135,10 +135,15 @@ function BaseInput({
           ? Number(event.target.value)
           : new Date(event.target.value);
       setLocalValue(newValue);
-      const boundedValue = boundsCheckedValue(newValue);
-      if (boundedValue !== value) debouncedOnChange(boundedValue);
+      const isValid = boundsCheckedValue(newValue);
+      if (isValid) {
+        debouncedOnChange(newValue);
+      } else {
+        // immediately send the last valid value to onChange
+        debouncedOnChange.flush();
+      }
     },
-    [boundsCheckedValue, value, debouncedOnChange]
+    [boundsCheckedValue, debouncedOnChange]
   );
 
   return (
--- a/src/components/widgets/NumberAndDateInputs.tsx
+++ b/src/components/widgets/NumberAndDateInputs.tsx
@@ -1,19 +1,20 @@
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect, useCallback, useMemo } from 'react';
 
 import { Typography, TextField } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import { DARK_GRAY, MEDIUM_GRAY } from '../../constants/colors';
 import { NumberOrDate } from '../../types/general';
+import { debounce } from 'lodash';
 
 type BaseProps<M extends NumberOrDate> = {
   /** Externally controlled value. */
-  value: M | '';
+  value?: M;
   /** Minimum allowed value (inclusive) */
   minValue?: M;
   /** Maximum allowed value (inclusive) */
   maxValue?: M;
   /** Function to invoke when value changes. */
-  onValueChange: (newValue: NumberOrDate | '') => void;
+  onValueChange: (newValue?: NumberOrDate) => void;
   /** UI Label for the widget. Optional */
   label?: string;
   /** Additional styles for component container. Optional. */
@@ -46,6 +47,17 @@ type BaseInputProps =
  * Input field taking a value we can do < > <= => comparisons with
  * i.e. number or date.
  * Not currently exported. But could be if needed.
+ *
+ * This component will allow out-of-range and empty values, but it will only
+ * call `onValueChange` when the new value is valid. An error message will be
+ * displayed when the value is invalid, but the consumer of this component will
+ * not be notified of the invalid state. It's possible we will want to add a
+ * callback to allow observing invalid states, in the future.
+ *
+ * The `onValueChange` callback is debounced at 500ms. This allows the user to
+ * type a value at a reasonable pace, without invoking the callback for
+ * intermediate values. We use a local state variable to track the input's
+ * actual value.
  */
 function BaseInput({
   value,
@@ -57,6 +69,7 @@ function BaseInput({
   containerStyles,
   displayRangeViolationWarnings = true,
 }: BaseInputProps) {
+  const [localValue, setLocalValue] = useState<NumberOrDate | undefined>(value);
   const [focused, setFocused] = useState(false);
   const [errorState, setErrorState] = useState({
     error: false,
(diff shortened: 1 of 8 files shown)


src/components/widgets/NumberAndDateInputs.tsx
```diff
@@ -83,7 +83,7 @@ function BaseInput({
     } else {
       setErrorState({ error: false, helperText: '' });
     }
-    return undefined;
+    return newValue;
```
bobular, line 86: Dave did this while making all input widgets 'controlled' - but I don't see why, so I reverted it here.  Still need to do lots of testing.

It may be the explanation for #66 - @moontrip

dmfalke: I did that to prevent calling `onChange` on the initial render. Perhaps instead we should update the `useEffect` hook just below and only call `onChange` if `newValue != null && newValue !== value`?

bobular: Sounds good - I'll have a play with that.

bobular:
New slider with auxiliary text-input box is working - it just needs the `showTextInput={true}` prop.

Lots of tidy-up was done in number and date inputs, and their stories - removing uncontrolled stories, making the controlled value prop mandatory and much more.

No known issues - other than perhaps the logic that automatically sets the width of the text box of the enhanced slider could be moved inside the NumerAndDateInputs.tsx code.  I'll add a comment to locate that code.

src/components/widgets/Slider.tsx
```diff
@@ -147,6 +154,24 @@ export default function SliderWidget({
...
+        <NumberInput
+          value={localValue}
+          minValue={minimum}
+          maxValue={maximum}
+          onValueChange={(newValue?: NumberOrDate) =>
+            handleChange(null, newValue as number)
+          }
+          displayRangeViolationWarnings={false}
+          containerStyles={{
+            width:
```
bobular, line 167: Here's the automatic width code.

Could possibly go into NumberInput and DateInput ?

src/components/widgets/NumberAndDateInputs.tsx
```diff
@@ -7,17 +7,19 @@ import { NumberOrDate } from '../../types/general';
 
 type BaseProps<M extends NumberOrDate> = {
   /** Externally controlled value. */
-  value?: M;
+  value: M;
```
dmfalke, line 10: Shouldn't we allow an empty value?

src/components/widgets/NumberAndDateInputs.tsx
```diff
@@ -7,17 +7,19 @@ import { NumberOrDate } from '../../types/general';
...
   /** Minimum allowed value (inclusive) */
   minValue?: M;
   /** Maximum allowed value (inclusive) */
   maxValue?: M;
   /** Function to invoke when value changes. */
-  onValueChange: (newValue: NumberOrDate | undefined) => void;
+  onValueChange: (newValue: NumberOrDate) => void;
```
dmfalke, line 16: This could be useful if we wanted to include a "clear value" button, or something similar.

src/components/widgets/NumberAndDateInputs.tsx
```diff
@@ -66,47 +69,49 @@ function BaseInput({
...
     const newValue = boundsCheckedValue(value);
-    if (newValue != null) onValueChange(newValue);
-  }, [minValue, maxValue]);
+    if (newValue !== value) onValueChange(newValue);
+  }, [boundsCheckedValue, value, onValueChange]);
 
-  const handleChange = (event: any) => {
-    if (event.target.value.length > 0) {
+  const handleChange = useCallback(
+    (event: any) => {
```
dmfalke, line 105:
```suggestion
    (event: React.ChangeEvent<HTMLInputElement>) => {
```

dmfalke (changes requested):
I've left some inline comments.

src/components/widgets/NumberAndDateInputs.tsx
bobular: See next comment

src/components/widgets/NumberAndDateInputs.tsx
bobular: I originally wanted to keep the ability to clear the inputs.  For one thing, it makes it easier to edit the contents if you can clear it before typing in the number you want.

However, there were problems...

React complained (in the console.log) about switching between controlled and uncontrolled components when the box is cleared. I think the problem is when the prop's value switches between defined and undefined.  I tried to use `null` instead but that didn't seem to help. The React warning suggested using empty string as a value, which I tried, e.g.

props
```
value: M | '';
onValueChange: (newValue: NumberOrDate | '') => void;
```

But on clearing the box it still went to zero, not empty.  (I'm pretty sure there was appropriate code in `boundsCheckedValue` to handle the empty string.)

I'll have another go though. Now the code is cleaned up a bit, I might be able to get it working more easily.

bobular: Oh...

`Number('') === 0`

bobular:
The text field inputs can now be cleared by the user, but...  there is still one issue

In this story `/story/widgets-number-input--silent-bounds-check` for example, you can't easily clear the input because a single backspace will create the value '1' which is outside the range, so is 'fixed' to 10.  The only way to clear it is to select everything in the box, and then press backspace.

I tried adding debounce logic similar to Slider.tsx but that doesn't work.  What's needed is debouncing *only* for the typing input, not the up/down arrow (or keyboard arrows).  

A non-debouncing solution would be to automatically select-all inside the input field when the user clicks in it.

moontrip:
@bobular 

> A non-debouncing solution would be to automatically select-all inside the input field when the user clicks in it.

I am not quite sure whether you already knew, but just in case, it can be achieved as follows:

`const handleFocus = (event: React.FocusEvent<HTMLInputElement>) => event.target.select();`

then add onFocus prop at the child component:

`onFocus={handleFocus}`

Actually I am adding it as a part of my current work, #66 adjusting selected range at NumberAndDateInputs.tsx component, so you may not need to make it if you plan to use it at the same file/component: I am using onFocus prop at TextField component in this case.

dmfalke:
Another thing to try in `NumberAndDateInputs`:
- Use component state for values, similar to `Slider`
- Allow invalid values
- Display an error message for invalid values
- Only call `props.onChange` for valid values
- Debounce the call to `props.onChange` to allow valid intermediate values without triggering a change

This should allow users to clear the input field before typing a new value. The range component wouldn't need to do this trick, since it would only call its `props.onChange` if the input calls its `props.onChange`.

If I have time tomorrow morning, I will try this out and report back.

bobular:
Thanks @moontrip - I didn't know that!

@dmfalke - that sounds worth a try - is `''` still a valid value for the externally controlled prop?

I'll have a go tonight if you don't find time today.

bobular:
Because I'm not working until around 2pm EST, it would be a good time to coordinate between this PR and DK's work for #66 - because it sounds like there is overlap.  So please feel free to do anything with this branch until that time.

dmfalke:
@bobular @moontrip I pushed a commit that implements my suggestion. I've tested it and it seems to work pretty well. Please take a look at your convenience.

bobular:
That looks great Dave - I have given it a thorough test and made one change so that the 
`/story/widgets-number-input--controlled-bounds` story works (changing the bounds causes the bounds checking to be done again)

I've noticed one remaining thing (sorry!)
I don't fully understand the logic behind the debounce cancel at L143 so would appreciate having that explained...

But I played around with and without that line canceled out.

As is: If you hold down on the up-arrow and let it repeat to an out-of-range value (e.g. 15 in `/story/widgets-number-input--controlled-bounded`) I notice that the debouncedOnChange never gets called and the externally controlled value is unchanged.

L143 (cancel) commented out: If you hold down up-arrow as described above, the external value gets set to 5 (which might be desired?)  But at some point that 5 gets transfered to localValue and so the displayed value goes down to 5 temporarily and then keeps incrementing upward (if the user is still holding down the up arrow).  So the behaviour with the controlled value is better (arguably) but the on-screen behaviour is a bit unnerving.

I'll see if I can figure out a simple solution, but we could discuss tomorrow.

bobular:
I've pushed a commit that (in my opinion) solves the problem of going out of range but not setting the controlled value to the limit of the range (as described above).

If you up-arrow out of range, when you release the key or mouse button, the controlled value will be set to 5, and after a 500ms warning, the displayed value will also be set to 5.  If you increase the displayed value again, it will not be "auto-corrected" back to 5 (because the controlled value of 5 has never changed).

I added a new story to show the behaviour when an externally provided value is out of range (nothing crazy happens - the out-of-range value is displayed with a warning).

I think this is a pretty OK solution all round.  It works nicely for the number-slider `/story/widgets-slider--auxiliary-text-input` - if you really want go out of range, you can, but your first attempt will be "corrected" and the controlled value will of course stay within the bounds.

dmfalke:
> I don't fully understand the logic behind the debounce cancel at L143 so would appreciate having that explained...

The idea here is that we don't want to notify the consumer of a stale value. Without this line, if `localValue` goes out-of-bounds or empty, a pending call to `onValueChange` will trigger the consumer (e.g., parent) of this component to update with an out-of-date value, which will cause the input's value to change "under the user's mouse". Thus, we have to cancel the pending call. Another option would be to call the `flush` method, which give you your desired behavior: the consumer will be notified of the last in-bounds value.

I haven't looked at your latest, but based on your description, I am concerned about changing the value like that. I will try to take a look... or maybe you can try my suggestion and compare the UX?

bobular:
Thanks for the explanation.

> Another option would be to call the `flush` method, which give you your desired behavior: the consumer will be notified of the last in-bounds value.
> 
> I haven't looked at your latest, but based on your description, I am concerned about changing the value like that. I will try to take a look... or maybe you can try my suggestion and compare the UX?

It does work, perfectly! Thanks again!

bobular (commented):
I can't approve the requested changes.  @dmfalke please could you?

dmfalke (approved):

PR #101, opened 2021-04-14
[issue-66] an approach for adjusting selected range

moontrip:
@dmfalke and @bobular Since #99  may solve this issue, #66, my work may not be needed anymore. Thus, this PR just intends to share what my approach was. There are couple of more things to do for completeness but I just stopped doing further owing to #99 . So, please just take a look at it and close this PR as this may not be worthwhile :)

Simply put, my approach was to move onChange function (handleChange) at NumberAndDateInputs component to the parent component, NumberAndDateRangeInputs (named, handleOnChange). This resolved the issue that selected range inputs could not be changed after mouse selection. One of disadvantages in this approach is that additional parameters/props are needed to distinguish min from max (or vice versa) as the onChange function is out of NumberAndDateInputs component: so, I used name property (for TextField/Input) and upperLabel & lowerLabel props to set the different value of the name property for min and max inputs respectively

### DIFF
--- a/src/components/plotControls/HistogramControls.tsx
+++ b/src/components/plotControls/HistogramControls.tsx
@@ -225,6 +225,9 @@ export default function HistogramControls({
           valueType !== undefined && valueType === 'date' ? (
             <DateRangeInput
               label="Selected Range"
+              // need to set lowerLabel and upperLabel for handleOnChange (onChange)
+              lowerLabel="selectedRangeMin"
+              upperLabel="selectedRangeMax"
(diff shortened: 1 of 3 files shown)


bobular:
Thanks for the PR @moontrip - I've made it a draft PR so it doesn't get merged accidentally.  Actually I'm going to close it too.

I've just tested web-eda/main with web-components/issue-94-number-slider (#99) and the selected range *is* now adjustable using the text boxes (though there are still some teething troubles).  The problem was changes in #76 which made all text fields un-editable - we should have made sure the stories were up to date and working.

PR #106, opened 2021-04-16
Change handling of dates from Date to string

bobular:
This is going to make io-ts validation of saved UI state much easier...


bobular:
I changed it to a full PR - but we can maybe wait to merge until I've tested it in full with the EDA, which I'm fixing now in a similar `date-to-string` branch.

bobular:
wrt date math:

For histogram, we need to send a bar width in milliseconds to Plotly.  This can't be calculated back-of-the-envelope style due to leap years, different length months etc.  Some kind of specialist date library is needed.

Likewise binMiddles used in blue-rectangle histogram selection are calculated using date math.

dmfalke (approved):

PR #116, opened 2021-05-05
Eliminate need for keys in CustomNode

dmfalke:
The need for keys in the render method of `CustomNode` can be eliminated by using a React `Fragment`, rather than an array. This requires a change to the `renderNode` prop of `EntityDiagram`.


chowington:
Looks good, thanks Dave!

PR #119, opened 2021-05-11
Selected Range and x-axis zoom interaction fixes

### LINKED ISSUE #117: Histogram: x-axis range selection should not impact selection-range bounds
I think the state represented in the screenshot should be legal.
image

bobular:
Associated EDA PR https://github.com/VEuPathDB/web-eda/pull/84 

Fixes #117 

Raised a possible back end issue
https://github.com/VEuPathDB/EdaDataService/issues/31

### DIFF
--- a/src/components/plotControls/HistogramControls.tsx
+++ b/src/components/plotControls/HistogramControls.tsx
@@ -27,6 +27,7 @@ import {
   NumberRangeInput,
   DateRangeInput,
 } from '../widgets/NumberAndDateRangeInputs';
+import LabelledGroup from '../widgets/LabelledGroup';
 
 /**
  * Props for histogram controls.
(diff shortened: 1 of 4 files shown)


bobular:
Sorted the issues without any need for back end changes.

dmfalke (commented):
@bobular, do you think we really need to bound the selected range? I don't really see an issue with the selection being outside the bounds of the data.

bobular:
> @bobular, do you think we really need to bound the selected range? I don't really see an issue with the selection being outside the bounds of the data.

We could remove the constraints, @dmfalke - but are they causing any issues at the moment?  The range bounds are used to fill in the min or max when only one number is entered (similar to live site behaviour) - I'll comment in this PR where that's happening - so that would have to be implemented differently.

bobular:
Hi @dmfalke and others.
I have done some refactoring (new LabelledGroup container component) and moved the selectedRange controls above the histogram.  It looks good there.  Seems that HistogramControls could lose some mandatory props but I'll leave that for another day.

I've done my best re. styling/layout (including darkening the lines and widening the widgets) - but it's not perfect.

dmfalke (approved):
This looks good to me!

PR #121, opened 2021-05-11
making Scatter plot controls

moontrip:
Implemented scatter plot controls that would be used at scatter plot viz to change data response (and plot). A story is made to test and usePlotControls hook is changed/utilized for that purpose. With the addition of of the "bestFitLineWithRaw" response, I feel that the "SmoothedMean" may be no longer required, but it is included anyway. A screenshot is attached here.

scatterplotcontrols1

### DIFF
--- a/src/components/plotControls/ScatterplotControls.tsx
+++ b/src/components/plotControls/ScatterplotControls.tsx
@@ -1,0 +1,122 @@
+import React, { useMemo } from 'react';
+import useDimensions from 'react-cool-dimensions';
+
+// Definitions
+import { LIGHT_BLUE, LIGHT_GRAY } from '../../constants/colors';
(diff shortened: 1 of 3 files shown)


dmfalke (commented):
This is looking pretty good. I was expecting to see three toggles for this: **raw**, **smoothed mean**, and **best fit**. The current implementation seems overwhelming, to me. Maybe we can get some feedback in our meeting today.

moontrip:
Thank you for your review @dmfalke. The button texts are the same with the responses' names from the backend, but perhaps we may consider shorter texts like your suggestion and then handle them at scatter plot viz code: I agree with your opinion about feedback today 👍

dmfalke (approved):

moontrip:
Thank you for your review, @dmfalke

PR #125, opened 2021-05-19
updates on scatter plot component and its plot control

moontrip:
several changes are made on scatter plot component and its plot control:
a) border and text of scatter plot control are removed
b) adjusted some space (minimized gaps between scatter plot and its control)
c) changed button title from Data to Plot options
d) button texts of plot control now have space among words: considered changing a CSS prop of `textTransform` but didn't do that as it was set at widget, not control. Note that corresponding changes are needed at scatter plot viz (will make a PR at web-eda after doing other stuffs as well) 
e) added title prop
f) tested to add two new props styleWidth and styleHeight which allows input string (like percent, e.g., 100%). Not quite sure if this will be used but added anyway


bobular (commented):
Same here - please remove commented lines before merging.

bobular:
Some more comments - 

1. What is `Scatterplot.tsx`? Does anything use it?  Can it be removed? (yes please, if poss!)

If `ScatterAndLinePlotGeneral.tsx` is the main scatterplot component, maybe let's rename it `Scatterplot.tsx`.
The long name `ScatterAndLinePlotGeneral` is already out of date, as it's not `ScatterAndBestFitAndLinePlotGeneral` ;-)

Also remove ScatterAndLinePlot.tsx and any associated stories.

2. In ScatterAndLinePlotGeneral.tsx
```
  styleWidth?: number | string;
  styleHeight?: number | string;
```
In other components we've called them `width` and `height` - and for some reason we've mostly used `number | string` instead of `CSSproperties['width']`.

But... all this needs standardising, and I'll be doing the groundwork for that standardisation very soon.  

But... it seems that `styleWidth` and `styleHeight` are not used anywhere at the moment.
What is happening to the
```
            width={1000}
            height={600}
```
props passed to `ScatterplotWithControls` at line 340 of ScatterplotVisualization.tsx ?
Seems they are not used and the Typescript is a bit loose.

moontrip:
@bobular Thank you for your review 👍 As for removing comments, yes I had a plan to clean things up before merging. 
For your other comments, please see my response below starting with **DK:** 

> Some more comments -
> 
> 1. What is `Scatterplot.tsx`? Does anything use it?  Can it be removed? (yes please, if poss!)
> 
> If `ScatterAndLinePlotGeneral.tsx` is the main scatterplot component, maybe let's rename it `Scatterplot.tsx`.
> The long name `ScatterAndLinePlotGeneral` is already out of date, as it's not `ScatterAndBestFitAndLinePlotGeneral` ;-)
> 
> Also remove ScatterAndLinePlot.tsx and any associated stories.

DK: yes it needs some adjustment. The `ScatterPlot.tsx` was, if I remember correctly, initialized by someone else with only support scatter plot. After several discussions and revisions on spec, the `ScatterAndLinePlotGeneral` component now handles three types of plots such as scatter, line, and density plots. This is the reason why corresponding vizs are all dependent on the `ScatterAndLinePlotGeneral` component. Without objections from others, I will make them simple as you suggested.


> 1. In ScatterAndLinePlotGeneral.tsx
> 
> ```
>   styleWidth?: number | string;
>   styleHeight?: number | string;
> ```
> 
> In other components we've called them `width` and `height` - and for some reason we've mostly used `number | string` instead of `CSSproperties['width']`.
> 
> But... all this needs standardising, and I'll be doing the groundwork for that standardisation very soon.
> 
> But... it seems that `styleWidth` and `styleHeight` are not used anywhere at the moment.
> What is happening to the
> 
> ```
>             width={1000}
>             height={600}
> ```
> 
> props passed to `ScatterplotWithControls` at line 340 of ScatterplotVisualization.tsx ?
> Seems they are not used and the Typescript is a bit loose.


**DK:** It is because your approach is different from mine. When I saw your histogram component, you assigned width and height as union type (number or string), while I always extended PlotProps from PlotlyPlot component. Actually, we can use `extend PlotProps` and change width (and height) : number to union type - I just didn't want to change PlotlyPlot component as it is commonly used for all plots. If those change will be occurred, then we don't need the props styleWidth/styleHeight. I will remove those props and corresponding changes.

moontrip:
@bobular I have removed the unnecessary props and corresponding ones and pushed a commit here. I will defer to change the component name in the (near) future because it also involves changes at viz and several PRs concerning scatter plot are involved in there. After settling them down, then I will revisit the task :)

bobular:
Hi @moontrip - I note that you say above that someone else started work on the Scatterplot component, so don't take anything personally...  In general we need to avoid any dependencies on plotly in our plot component implementations (I notice I have a dependency in my Boxplot.tsx - bad Bob...). 

The worst offender in ScatterAndLinePlotGeneral.tsx is the `ScatterplotProps.data` prop, which depends on plotly's `PlotData`.  The data shape should be defined using non-plotly types in ScatterAndLinePlotGeneral.tsx or possibly in `../types/plot/*.ts` (as was done for Histogram).  You seem to be halfway to doing this correctly in Barplot.tsx (though at present it's `data: any`).

moontrip:
@bobular Thank you for sharing your thoughts and comments 👍 I think that your speaking is specifically related to the typing of `data` props and it is well accepted - I realized that I forgot to set data props at barplot! It was a simple one so I just updated barplot component.

The scatter plot component was my first plot component work so it was slipped for the configuration of data props more specifically. I will also see if I can do that today.

moontrip:
@bobular I have added data prop to the scatter plot component

src/plots/ScatterAndLinePlotGeneral.tsx
```diff
@@ -1,14 +1,29 @@
...
-type PlotDataKey = keyof PlotData;
+import { Layout } from 'plotly.js';
 
 // change interface a bit more: this could avoid error on data type
-export interface ScatterplotProps<T extends keyof PlotData> extends PlotProps {
-  data: Pick<PlotData, T>[];
+// export interface ScatterplotProps<T extends keyof PlotData> extends PlotProps {
+export interface ScatterplotProps extends PlotProps {
+  // data: any,
+  data: Array<{
```
bobular, line 9: Now the dependency on plotly for this data prop is gone, we have to provide more documentation, and I suggest some basic sanity checking.

For documentation please could you add comments on the line preceding every property in the /** blah blah */ format used elsewhere.  Can `mode`, `fill` and `line.shape` really just be a string? (They are enumerated types, aren't they?)  What units are the widths and sizes in? 

For sanity checking, I would want to check that x.length === y.length, but I see we haven't done that kind of thing in any of the other components.  Something to discuss at tomorrow's meeting.

bobular (commented):
Left a comment about documenting the props.

Reminder about the rationalisaion of all the Scatter*.tsx files...

src/plots/ScatterAndLinePlotGeneral.tsx
moontrip: Thank you for your detailed comments @bobular 
I agree with your suggestions about mode, fill and line.shape as they are enumerated types. Perhaps we can have limited number of possible list for our purpose: plotly would have more options. Also will add some description per prop as you suggested.

For checking x.length === y.length, actually those were already implemented at both scatter plot story and scatter plot viz where plotly friendly scatter plot data are generated. 

For the last one, scatter plot props, yes it is a bit complicated as the scatter plot component handles several plots at once and the way plotly demands for each plot stype (marker, line, confidence interval, density, and combinations of plots like marker + line + confidence interval). Will try to describe it at each prop (e.g., for scatter plot, for line plot, etc.)

moontrip: @bobular corresponding updates are made [here](https://github.com/VEuPathDB/web-components/pull/125/commits/fd97ad7d1ea3bb12e1810ef281860d073f08dd17)

moontrip:
tested empty data case and it works by sending empty array at data props

moontrip:
ScatterAndLinePlot and corresponding story are removed. ScatterAndLinePlotGeneral and corresponding story are renamed as XYPlot. Story file is changed to use the new name.

bobular (approved):
Hi, I'm approving this so we can get it merged ASAP, but please also consider doing the following before merging:

Remove:
```
src/stories/_ScatterPlot.stories.tsx
src/plots/ScatterPlot.tsx
src/stories/_LinePlot.stories.tsx
src/plots/LinePlot.tsx
```

Rename the following to XYPlot*
```
src/components/plotControls/ScatterplotControls.tsx
src/stories/plotControls/ScatterplotControls.stories.tsx
```
If we are planning to add axis range/zoom controls, then you may also want to adapt XYPlot to use our `NumberOrDateRange` type (see below).  It would probably be easiest to add a `data.type : 'number' | 'date'` property like in Histogram.tsx, and have XYPlot.data take `number[] | string[]`.  I started out with Histogram.tsx using the built in `Date` type but changed it back to use string. I think maybe best to add a new ticket for this ("Add range controls to XYPlot and use string instead of Date type").

```
  /** x-axis range: required for confidence interval */
  independentAxisRange?: NumberOrDateRange; // was number[] | Date[];
  /** y-axis range: required for confidence interval */
  dependentAxisRange?: NumberOrDateRange; // was number[] | Date[];
  /** show plot legend */
```

moontrip:
Thanks @bobular for your review! 👍 Please see my comments for the first two items in the following (starting with "DK:")

> Hi, I'm approving this so we can get it merged ASAP, but please also consider doing the following before merging:
> 
> Remove:
> 
> ```
> src/stories/_ScatterPlot.stories.tsx
> src/plots/ScatterPlot.tsx
> src/stories/_LinePlot.stories.tsx
> src/plots/LinePlot.tsx
> ```

**DK:** I may be wrong but I thought we would leave them as are per @dmfalke comment in the meeting like "there may be a use case to use simple scatter or line plot"? If both of you agree with the removal, then I will do that.


> Rename the following to XYPlot*
> 
> ```
> src/components/plotControls/ScatterplotControls.tsx
> src/stories/plotControls/ScatterplotControls.stories.tsx
> ```
> 

**DK:** I just leave the as are because of two reasons: a) 'scatterplot' itself is quite a unique name used for both the frontend and backend in viz; b) also in viz, we still use the name, ScatterplotVisualization. I am not quite sure if we change all of scatterplot related names to XYPlot. 


> If we are planning to add axis range/zoom controls, then you may also want to adapt XYPlot to use our `NumberOrDateRange` type (see below). It would probably be easiest to add a `data.type : 'number' | 'date'` property like in Histogram.tsx, and have XYPlot.data take `number[] | string[]`. I started out with Histogram.tsx using the built in `Date` type but changed it back to use string. I think maybe best to add a new ticket for this ("Add range controls to XYPlot and use string instead of Date type").
> 
> ```
>   /** x-axis range: required for confidence interval */
>   independentAxisRange?: NumberOrDateRange; // was number[] | Date[];
>   /** y-axis range: required for confidence interval */
>   dependentAxisRange?: NumberOrDateRange; // was number[] | Date[];
>   /** show plot legend */
> ```

**DK:** I agree with your opinion. The data/string stuff needs another ticket to be handled, not this one, because it would not be that simple task, I suppose

dmfalke:
@moontrip we decided to just have the one plot and name it `XYPlot` as @bobular said.

dmfalke:
> > Rename the following to XYPlot*
> > ```
> > src/components/plotControls/ScatterplotControls.tsx
> > src/stories/plotControls/ScatterplotControls.stories.tsx
> > ```
> 
> **DK:** I just leave the as are because of two reasons: a) 'scatterplot' itself is quite a unique name used for both the frontend and backend in viz; b) also in viz, we still use the name, ScatterplotVisualization. I am not quite sure if we change all of scatterplot related names to XYPlot but at least for web-component repo, it may be better to be consistent as you suggested.

I agree with this change, as well. I think we should keep the name **ScatterplotVisualization** over in web-eda, or, at the very least, consider that separately.

moontrip:
Thank you for sharing your thought, @dmfalke 👍

bobular:
Hi @moontrip - I think it's easier for developers if the stories are called ComponentName.stories.tsx, so please rename Scatterplot.stories.tsx 

> DK: I may be wrong but I thought we would leave them as are per @dmfalke comment in the meeting like "there may be a use case to use simple scatter or line plot"? If both of you agree with the removal, then I will do that.

Yes we did discuss having a `<Lineplot>` component which was a simplified wrapper around `<XYPlot>` with restricted props.  If you want to implement that, then please do (and create a Lineplot.stories.tsx (without the leading underscore).  You could do the same for Scatterplot and DensityPlot.  But these are not urgent and if you decide not to, please make a ticket "Simplified Lineplot, Scatterplot, etc wrappers for XYPlot" so we don't lose track of our thinking.

Likewise it would be nice, but not urgent, to add _simple_ scatter plot, line plot, etc stories to XYPlot.stories.tsx - by simple I mean hardcoded data with 3-5 data points.  If not doing now, please make a ticket "Add simpler use case stories to XYPlot.stories.tsx"

PR #128, opened 2021-05-21
Potential new plot control approach

bobular:
The `Plots -> Super Scatter` story in SuperScatter.stories.tsx has a scatterplot with a histogram for each variable.

There's a new `AxisControls` component that can handle
- log scale on/off
- axis range (for zooming)
- reset axis

This is used twice. (I **temporarily** added a trivial client-side x-axis zooming to Histogram (independentAxisRange) just to avoid having to use any dynamic data generation.)

I imagine some other generic controls such as `MarkerControls`:
- opacity
- stacked or grouped
- relative or absolute
These apply to a whole plot and not a specific axis (that is potentially debatable)

Also `RangeSelectionControls` and `BinningControls`.

The idea is to avoid duplication wrt HistogramControls, BarplotControls, BoxplotControls, PieControls, etc, and be ready to handle more complex visualisations.

image

### DIFF
--- a/src/components/plotControls/AxisControls.tsx
+++ b/src/components/plotControls/AxisControls.tsx
@@ -1,0 +1,87 @@
+import React from 'react';
+import {
+  NumberOrDateRange,
+  NumberRange,
+  DateRange,
+} from '../../../lib/types/general';
+import LabelledGroup from '../widgets/LabelledGroup';
+import {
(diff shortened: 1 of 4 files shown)


bobular:
Closing this to avoid confusion.

The general approach proposed here is still valid: roll your own controls in client code, and create reusable control components here as needed.

PR #133, opened 2021-05-27
Entity diagram counts and highlighting

dmfalke:
This PR includes support for adding counts and a filter icon to entity nodes.


bobular (approved):
I wonder if all the entity names should be in bold?

Otherwise all looks great.

dmfalke:
Thanks, @bobular. The reason to only bold the selected entity name is to make it stand out more. We can revisit later.

jtlong3rd (approved):

PR #138, opened 2021-06-02
box plot component for EDA

moontrip:
Unlike the pre-existing box plot component (by Bob) that include the computation of quartiles etc., Plotly.js simply demands raw data for box plot and those box plot parameters such as quartiles etc. are automatically handled at Plotly.js. Thus, I have made a new simpler box plot component for EDA, temporarily named as boxplotEDA to distinguish from pre-existing one. 

This is functional, but just made it as draft PR as this is probably not included in the Gates demo


src/plots/BoxplotEDA.tsx
```diff
@@ -0,0 +1,176 @@
+import { useEffect, useMemo, useState } from 'react';
```
bobular, line 1: Can we remove this source file before merging please?

src/stories/plots/Boxplot.stories.tsx
```diff
@@ -1,11 +1,9 @@
 import React from 'react';
 import { Meta, Story } from '@storybook/react';
 import Boxplot, { Props } from '../../plots/Boxplot';
-import stats from 'stats-lite';
-import _ from 'lodash';
 
 export default {
-  title: 'Plots/Box',
+  title: 'Plots/Boxplot',
```
bobular, line 6: Coordinating with #150 and before merging, please make sure all stories are "storybook control" enabled.  Thanks.

bobular (commented):
Looks good - just a few style and organisational issues.

Will look at the web-eda PR ASAP next.

src/plots/Boxplot.tsx
```diff
@@ -110,6 +110,20 @@ export default function Boxplot({
                 : undefined,
           };
 
+    // seems like plotly bug: y[0] or x[0] should not be empty array (e.g., with overlay variable)
```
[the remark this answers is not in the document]
moontrip: @bobular well actually I found that kind of bug when using overlay variable. I believe that in most cases, there would be no bug but if using overlay variable, there may be the case. I just checked your case, country vs water trips weekely but y[0] was not empty (actually `outliers` response) - at GEMS1 case control.

The test case is at its story file, Multiple Dataset, which Kenya has the outliers with empty array at the first array (y[0])

finally, I tried various way, but only y[0] = [null] worked so...

src/plots/Boxplot.tsx
```diff
@@ -110,6 +110,20 @@ export default function Boxplot({
...
+    if (
+      orientation === 'vertical' &&
+      orientationDependentProps.y &&
+      orientationDependentProps.y[0].length === 0
+    )
+      orientationDependentProps.y[0] = [null];
+    if (
```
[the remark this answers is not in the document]
moontrip: yep will do that

src/plots/Boxplot.tsx
```diff
@@ -122,3 +218,12 @@ Boxplot.defaultProps = {
   markerOpacity: 0.5,
   orientation: 'vertical',
 };
+
+function isArrayOfNumbers(value: any): value is number[] {
```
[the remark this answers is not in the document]
moontrip: wow I didn't know that. It was used to find and test the aforementioned bug but forgot to remove.

src/stories/plots/Boxplot.stories.tsx
moontrip: Well actually I have a plan to merge your changes in the future after your PR will be merged into the main. 
storybook control is already in there, `With Storybook Control` story

src/plots/BoxplotEDA.tsx
moontrip: Certainly I have a plan to remove them, BoxplotEDA and its story file. Will do that this time

src/plots/Boxplot.tsx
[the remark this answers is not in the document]
bobular: (My GEMS example does create `y[0] == []` - after subsetting on countries so the first country is Mozambique.)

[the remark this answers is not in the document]
moontrip: Ah, I see. I didn't use filter/subsetting so that might be the reason why I could not get your case 👍

moontrip:
updated box plot components for tidy-up and corresponding viz. Perhaps this can be out of draft PR so I am changing it

src/stories/plots/Boxplot.stories.tsx
```diff
@@ -19,68 +17,257 @@ export default {
...
+// I think the enrollment year's type should be date (classified as number in data though)
+const yearData = {
+  series: [
+    {
+      lowerfence: [2007, 2007, 2007],
+      upperfence: [2011, 2011, 2011],
+      q1: [2008, 2008, 2008],
+      q3: [2010, 2010, 2010],
+      median: [2009, 2009, 2009],
+      mean: [2008.8639, 2009.0816, 2009.0021],
```
bobular, line 167: These aren't dates!
I suggest something like this 
```
      mean: ['2008-10-02', '2009-07-22', '2009-01-13'],
```

src/plots/Boxplot.tsx
```diff
@@ -82,25 +109,41 @@ export default function Boxplot(props: BoxplotProps) {
     };
   });
 
+  console.log('data =', data);
```
bobular, line 112: console log

src/plots/Boxplot.tsx
```diff
@@ -37,40 +51,53 @@ export default function Boxplot(props: BoxplotProps) {
...
-        : {
-            y0: d.label,
-            x:
-              d.rawData && showRawData
-                ? [d.rawData]
-                : d.outliers.length
-                ? [d.outliers]
-                : undefined,
-          };
+    const [independentAxis, dependentAxis] =
```
bobular, line 64: Does it help (or hinder? perhaps eating memory for large data) to use `useMemo` for `const data = ...` ?

src/plots/Boxplot.tsx
```diff
@@ -16,14 +17,27 @@ export interface BoxplotProps
...
+  showDependentAxisTickLabel?: boolean;
+  /** independentValueType: 'string' | 'number' | 'date' | 'longitude' | 'category' */
+  independentValueType?:
+    | 'string'
+    | 'number'
+    | 'date'
+    | 'longitude'
+    | 'category';
+  /** dependentValueType: 'string' | 'number' | 'date' | 'longitude' | 'category' */
+  dependentValueType?: 'string' | 'number' | 'date' | 'longitude' | 'category';
```
bobular, line 40: Seeing `'longitude'` and `'category'` here makes me worry this is too EDA-centric.

Only the `'date'` value is used/implemented and tested with a story (dependentAxis dates for "Year Dataset" story).  (It seems impossible to select a date variable as dependentVariable in the EDA boxplot viz...?)

Also, there is no use-case for having the `independentValueType` prop, is there?  Won't the independent variable always be a categorical variable?   The independent axis needs to be handled the same way as barplot, I think.  Barplot.tsx doesn't (currently) specify `layout.xaxis.type`.

I suggest remove the `independentValueType` prop because, like Barplot at the moment, every compatible variable seems to be categorical-string.

`dependentValueType` should just have `'number' | 'date'` options until other options are required in the future.

If we need to do something special with categorical number variables in the future, we can come back and revisit.

src/plots/Boxplot.tsx
```diff
@@ -82,25 +109,41 @@ export default function Boxplot(props: BoxplotProps) {
...
+      type: dependentValueType === 'date' ? 'date' : undefined,
     },
-    [independentAxis]: {
-      title: independentAxisLabel,
-      range: data.length ? undefined : [1, 5], // avoids x==0 line
-      showgrid: false,
-      tickfont: data.length ? {} : { color: 'transparent' },
-    },
-    showlegend: false,
+    //DKDK don't forget this for multiple datasets
```
bobular, line 144: DKDK...

src/types/plots/boxplot.ts
```diff
@@ -1,23 +1,26 @@
...
+  /** upper whisker/fence optional */
+  upperfence?: number[];
   /** lower quartile (bottom of box) */
-  q1: number | string;
+  q1: number[];
+  /** upper quartile (top of box) */
+  q3: number[];
   /** median (middle line of box) */
-  median: number | string;
+  median: number[];
```
bobular, line 13: So if we are removing support for date-based dependent variables (which is probably fine, but we need to clarify with the backend team) - then we can also remove the `dependentValueType` prop, can't we?

I see now that the removal of `string[]` types here make my `yearData` story example give a TS error.

TLDR: either offer full date support for dependentAxis or none at all!

src/types/plots/boxplot.ts
```diff
@@ -1,23 +1,26 @@
...
+  label: string[];
+  /** legend name */
+  name?: string;
   /** color for this box */
   color?: string;
   /** optional complete data (not recommended for huge datasets) */
-  rawData?: number[] | string[];
-  /** outliers (data points outside upper and lower whiskers/fences */
-  outliers: number[] | string[];
+  rawData?: number[][] | string[][];
```
bobular, line 23: Why is `string[][]` still here if it was removed for mean/median/q1/q3 etc?

bobular (changes requested):
Hi DK - good job accommodating the tidy-up changes.

There are some major decisions to take about handling date data, which will need to be resolved before tidying up this ticket and merging.

In the meantime, you could remove a few DKDKs ;-) 

Note that I made this review *before* looking at the sister PR in web-eda

src/plots/Boxplot.tsx
```diff
@@ -37,40 +51,53 @@ export default function Boxplot(props: BoxplotProps) {
...
+        d.rawData && showRawData
+          ? d.rawData
+          : d.outliers?.length
+          ? d.outliers
+          : undefined,
+    };
+
+    // seems like plotly bug: y[0] or x[0] should not be empty array (e.g., with overlay variable)
+    // see multipleData at story file (Kenya case)
+    if (
```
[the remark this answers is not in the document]
moontrip: @bobular thank you for your comments and suggestions. I will check this part. DKDK and console.log would be cleaned up in the final stage when all stuffs will be approved :)

src/plots/Boxplot.tsx
```diff
@@ -16,14 +17,27 @@ export interface BoxplotProps
...
   /** show the rawData (if given) - optional */
   showRawData?: boolean;
   /** Show the mean as an extra dotted line in the box - optional */
   showMean?: boolean;
+  /** show/hide independent axis tick label */
+  showIndependentAxisTickLabel?: boolean;
+  /** show/hide dependent axis tick label */
+  showDependentAxisTickLabel?: boolean;
+  /** independentValueType: 'string' | 'number' | 'date' | 'longitude' | 'category' */
+  independentValueType?:
```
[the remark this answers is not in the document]
moontrip: @bobular I have separate them to be simpler for the use of typing at its viz: certainly I may define data part separately like histogram. In my opinion, we may separate them as they are not a part of Data structure.

src/stories/plots/Boxplot.stories.tsx
moontrip: @bobular as indicated at the story file, those are actual response data format for that case. I just tested with Gem1 study and in that case there was no date data - other study may have date data though: any case (X/Y variables and study) you can suggest testing at viz ?

src/plots/Boxplot.tsx
moontrip: @bobular yep I will do that later

src/plots/Boxplot.tsx
moontrip: @bobular not sure if useMemo will be much helpful in this simple case.

src/plots/Boxplot.tsx
moontrip: @bobular Initially I just set number and date, however, I had to change it because of type errors at viz, which they suggest using those types. I don't think it would hurt anyway. Yeah I agree that `independentValueType` is an overkill at this stage.

src/plots/Boxplot.tsx
moontrip: @bobular yep I will get rid of them later 😄

src/types/plots/boxplot.ts
moontrip: @bobular I just followed the [data API spec](https://veupathdb.github.io/EdaDataService/api.html#resources:/apps/pass/visualizations/boxplot:post) as median is currently listed as number[]. It could be incorrect so we need to make sure of it at first.

src/types/plots/boxplot.ts
moontrip: @bobular The data API doc did not clarify the type of the rawData so I just left it as number or string. Again, I tried to follow the data API doc for defining all response data types.

src/plots/Boxplot.tsx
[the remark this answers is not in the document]
moontrip: @bobular updated this with your suggestion 👍

src/plots/Boxplot.tsx
moontrip: @bobular removed this

src/plots/Boxplot.tsx
moontrip: @bobular removed  `independentValueType` prop (and at viz too)

src/plots/Boxplot.tsx
moontrip: @bobular removed all DKDK 😃

moontrip:
@bobular I made comment at each suggestion if I addressed them. The others are debatable so I just leave them as is

src/stories/plots/Boxplot.stories.tsx
bobular: Oh I see - well that's a data problem - it's a date stored as a number.  Not our problem.  The story should use real dates though (or we remove all date implementation from the plot component, see below)

There are plenty of real dates (stored as dates) in GEMS1 - but the variable constraints do not allow these to be chosen in the EDA for boxplot y-axis.  I assume this means we will not implement it, but I will ask on the backend Slack for confirmation.

moontrip: That makes sense. I could not see any date-like variables when selecting y-axis variable so suspected as well it did not allow date by the constraints. Thank you for handling that possible backend issue 👍

src/plots/Boxplot.tsx
bobular: I can't see why dependentValueType needs all those options. The type error you see is on the viz side because you're passing `variable.type` which is typed as `'string' | 'number' | 'date' | 'longitude' | 'category'` into the Boxplot component.

I don't think that's a good enough reason to copy those types into Plot Components.  The tail shouldn't wag the dog!

src/types/plots/boxplot.ts
bobular: Ah yes, the API spec says no dates (so does Nupur on #eda-backend)

src/types/plots/boxplot.ts
bobular: Looks like we should remove the string options here - based on what we have just learned.

src/plots/Boxplot.tsx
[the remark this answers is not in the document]
bobular: I happy to agree with you

src/plots/Boxplot.tsx
moontrip: Yes you are right as I am sending variable.type directly. Certainly I can also send only 'number' or 'date' value from viz by checking variable.type with condition. But I think it is better to just pass variable.type value without such an additional check/computing as it is more generic ? 

`dependentValueType` at plot component is just used for assigning `type` parameter at plotly's axis, `date | linear/undefined`,  and actually it only matters when dependentValueType = 'date', so it is not that harmful to have other values beyond number and date ?

If you still think it should be `number | date` only then I can change it :)

src/types/plots/boxplot.ts
moontrip: yeah I will remove string[][] in there (for both rawData and outliers)

src/plots/Boxplot.tsx
moontrip: since no 'date' data is expected at boxplot, perhaps we can either remove that props or just leave it as number| date - I will work for the latter

bobular:
I think we have two choices

1.
Keep the dependentValueType prop:
'number' | 'date' and defaults to 'number'

Fix the BoxplotData to allow number or string everywhere.

2.
Remove all date handling.  EDA doesn't want it, and if we need it for
another client later, we can add it.

bobular (approved):

bobular:
Sorry for the delay - I didn't see this had gone stale.  I just approved the changes!

jtlong3rd (approved):

PR #142, opened 2021-06-04
Add loading spinner

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/88

### DIFF
--- a/src/plots/MosaicPlot.tsx
+++ b/src/plots/MosaicPlot.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PlotlyPlot, { PlotProps, ModebarDefault } from './PlotlyPlot';
 import { PlotParams } from 'react-plotly.js';
+import CircularProgress from '@material-ui/core/CircularProgress';
 import _ from 'lodash';
 
 export interface Props extends Omit<PlotProps, 'width' | 'height'> {
@@ -18,6 +19,7 @@ export interface Props extends Omit<PlotProps, 'width' | 'height'> {
   titleSize?: number;
   width?: number | string;
   height?: number | string;
+  showSpinner?: boolean;
 }
 
 export default function MosaicPlot(props: Props) {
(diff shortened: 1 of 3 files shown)


chowington:
Works for mosaic! Still need to add for the other graphs

src/plots/MosaicPlot.tsx
```diff
@@ -18,6 +19,7 @@ export interface Props extends Omit<PlotProps, 'width' | 'height'> {
   titleSize?: number;
   width?: number | string;
   height?: number | string;
+  showSpinner?: boolean;
```
chowington, line 22: Oops, don't need this

chowington:
Added for all graphs! We should work on a way to add features to all graphs without having to add it individually to each. A bit of refactoring is in order

src/plots/XYPlot.tsx
```diff
@@ -100,21 +101,35 @@ export default function XYPlot(props: ScatterplotProps) {
...
+            showlegend: props.displayLegend,
+          },
+        }}
+        config={{
+          displayModeBar: props.displayLibraryControls ? 'hover' : false,
+          staticPlot: props.staticPlot,
+        }}
+      />
+      {props.showSpinner && (
+        <div
```
jtlong3rd, line 122: To keep things DRY, let's move this TSX into some sort of reusable `LoadingSpinner` component.

You might also try eliminating the duplicate `<div style={{ position: 'relative' }}>` and `props.showSpinner &&` by making some sort of `ReloadableContainer` component which receives `children: ReactNode` and `showSpinner?: boolean` props, and renders the following TSX:

```
<div style={{ position: 'relative' }}>
  {props.children}
  {props.showSpinner && <LoadingSpinner />}
</div>
```

jtlong3rd (changes requested):
I like the new spinner!

I agree that it can be tedious to add new "features" to all our graphs, and that in the long term, we could think about a more ergonomic design. For right now, I had one small suggestion to keep the current implementation DRY.

bobular (commented):
I wonder if it can even be done at the PlotlyPlot.tsx level?  Integrate it somehow alongside the Suspense lazy loading handling (and use the same spinner and use the height and width from PlotProps)...?

This may depend on the big tidy-up (for example, currently Histogram's props do not extend PlotlyPlot's PlotProps... - most likely my bad) 

Should we merge this and flag it for the tidy-up?

src/plots/XYPlot.tsx
jtlong3rd: Feel free to address my / @bobular 's suggestions in a followup issue!

jtlong3rd (approved):

chowington:
Thanks @jtlong3rd! I'll work on the DRYing you suggested!

@bobular It's not currently so simple as adding something purely to `PlotProps` because `PlotProps` props have to get handled individually by each plot currently. Each plot has to transform its own unique props (that may or may not extend `PlotProps`) to Plotly's default `PlotParams`, as that's what `PlotlyPlot` takes. Hopefully that makes sense!

dmfalke:
I do agree with @bobular that we can probably add the spinner itself to the implementation of `PlotlyPlot`. It would just need to have a new `showSpinner` prop.

In terms of DRYing up the props interfaces, it might make sense to have a shared interface that all plots extend. Note that this would be different from `PlotProps` currently used by `PlotlyPlot` (we should probably rename that to `PlotlyPlotProps`).

PR #143, opened 2021-06-05
new radio widget, stories, changes for controls

moontrip:
This is required for the change from box button to radio button for scatterplot viz. This PR includes several works:

- created radio button group widget with various props and made its story
- changed XYPlotControls with the new widget and its story
- changed useControls with the name of XYPlot
- clean-up barplot and its story

A screenshot is attached:

radiowidget1

### DIFF
--- a/src/components/plotControls/XYPlotControls.tsx
+++ b/src/components/plotControls/XYPlotControls.tsx
@@ -38,7 +38,7 @@ export type XYPlotControlsProps = {
   /** button color: for now, supporting blue and red only - primary: blue; secondary: red */
   buttonColor?: 'primary' | 'secondary';
   /** margin of radio button group: string array for top, left, bottom, and left, e.g., ['10em', '0', '0', '10em'] */
-  setMargin?: string[];
+  margins?: string[];
 };
 
 /**
@@ -60,7 +60,7 @@ export default function XYPlotControls({
   labelPlacement,
   minWidth,
   buttonColor,
-  setMargin,
+  margins,
 }: XYPlotControlsProps) {
   const { ref, width } = useDimensions<HTMLDivElement>();
 
@@ -113,7 +113,7 @@ export default function XYPlotControls({
             labelPlacement={labelPlacement}
             minWidth={minWidth}
             buttonColor={buttonColor}
-            setMargin={setMargin}
+            margins={margins}
           />
         )}
       </div>
--- a/src/components/plotControls/XYPlotControls.tsx
+++ b/src/components/plotControls/XYPlotControls.tsx
@@ -4,14 +4,13 @@ import useDimensions from 'react-cool-dimensions';
 // Definitions
 import { LIGHT_BLUE, LIGHT_GRAY } from '../../constants/colors';
 import { ErrorManagement } from '../../types/general';
-import ControlsHeader from '../typography/ControlsHeader';
 
 // Local Components
-import ButtonGroup from '../widgets/ButtonGroup';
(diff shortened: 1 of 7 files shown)


moontrip:
a new commit is added to allow barplot's width/height props to accept both number and height values. This will be used at barplot viz as well

src/components/widgets/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,115 @@
...
+  labelPlacement,
+  minWidth,
+  buttonColor,
+  setMargin,
+}: RadioButtonGroupProps) {
+  // perhaps not using focused?
+  // const [focused, setFocused] = useState(false);
+
+  //DKDK
+  console.log('selectedOption = ', selectedOption);
```
jtlong3rd, line 53: Can we delete these `console.log`s?

PS: Out of curiosity, have you tried [React DevTools](https://www.npmjs.com/package/react-devtools)? I find it very helpful for keeping track of prop, state, and other hook values.

src/components/plotControls/XYPlotControls.tsx
```diff
@@ -24,16 +23,26 @@ export type XYPlotControlsProps = {
...
+  /** How buttons are displayed. Vertical or Horizontal */
+  orientation?: 'vertical' | 'horizontal';
+  /** location of radio button label: start: label & button; end: button & label */
+  labelPlacement?: 'start' | 'end' | 'top' | 'bottom';
+  /** minimum width to set up equivalently spaced width per item */
+  minWidth?: number;
+  /** button color: for now, supporting blue and red only - primary: blue; secondary: red */
+  buttonColor?: 'primary' | 'secondary';
+  /** margin of radio button group: string array for top, left, bottom, and left, e.g., ['10em', '0', '0', '10em'] */
+  setMargin?: string[];
```
jtlong3rd, line 41: I would name this prop something like `margins`. The name `setMargin` would make more sense if the prop were some sort of callback.

jtlong3rd (changes requested):
This looks great! I had two minor suggestions.

src/components/widgets/RadioButtonGroup.tsx
moontrip: @jtlong3rd First of all, thank you for your review 👍 For that console.log(), usually I tend to clean up those stuffs after approval, as they may need more works after review, but anyway I will remove them.

Yes, I am using react devtools definitely, but yeah it is a bit old habit to use console.log() :)

src/components/plotControls/XYPlotControls.tsx
moontrip: @jtlong3rd Actually I was gauging how to name that props: `margin/margins/setMargin/setMargins` haha. Certainly `setMargin` is more like callback stuff as you said, so will change it to `margins`

src/components/widgets/RadioButtonGroup.tsx
jtlong3rd: No worries - it's an old habit for me too! 😆

bobular (approved):
Approve, after taking into consideration Jamie and my suggestions about the margin prop.

Thanks

jtlong3rd (approved):

moontrip:
I have discussed with Bob regarding the spacing one. We agree that there needs to be major tidy-up of all the plot props, so for the time being I will use margins props as is and then revisit with more standardized props later. Will merge this.

Thank you for your reviews, @bobular and @jtlong3rd 👍

PR #146, opened 2021-06-08
add a default prop, fixedrange, at PlotlyPlot

moontrip:
this will address the issue at web-eda, [All visualization plots need static axes (undraggable/unscrollable)](https://github.com/VEuPathDB/web-eda/issues/143)

### DIFF
--- a/src/plots/Histogram.tsx
+++ b/src/plots/Histogram.tsx
@@ -350,7 +350,6 @@ export default function Histogram({
     },
     color: textColor,
     range: [minBinStart, maxBinEnd],
-    fixedrange: true,
     tickfont: data.series.length ? {} : { color: 'transparent' },
   };
   const dependentAxisLayout: Layout['yaxis'] | Layout['xaxis'] = {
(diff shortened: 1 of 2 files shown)


bobular:
Looks good. Does it matter if a plot doesn't have an x-axis or y-axis? (does PiePlot complain?)

moontrip:
> Looks good. Does it matter if a plot doesn't have an x-axis or y-axis? (does PiePlot complain?)

@bobular I tried to use pieplot/donut stories and it seems to be fine without complaints :)

bobular (changes requested):
I basically approve, but please remove the `fixedrange: true` from Histogram.tsx first.

Also let's discuss tomorrow if we really never want to have Plotly's dynamic axes (e.g. in a non-EDA setting).

moontrip:
@bobular I just removed the fixedrange options from histogram component

jtlong3rd (approved):

bobular (approved):

dmfalke (approved):

PR #147, opened 2021-06-08
remove SmoothedMean option in XYPlotControls

moontrip:
One of the XYPlotControls option, `SmoothedMean`, is now removed from backend, thus it is also removed at the plot control

Here is a screenshot. Although all options are equally spaced per request by Danica etc., I feel that the space between the `Raw` and `Smoothed Mean With Raw` may be better to be narrowed (or wider), especially without the SmoothedMean option. This may be alleviated with the props, `minWidth`, though

scatter-control1

### DIFF
--- a/src/components/plotControls/XYPlotControls.tsx
+++ b/src/components/plotControls/XYPlotControls.tsx
@@ -102,7 +102,6 @@ export default function XYPlotControls({
             label="Plot options"
             options={[
               'Raw',
-              'Smoothed mean',
               'Smoothed mean with raw',
               'Best fit line with raw',
             ]}


jtlong3rd (approved):

dmfalke (approved):
I agree that equal-width labels doesn't look that great, but I'm okay with moving ahead with this.

moontrip:
I think this can be merged without Bob's review/approval as this is quite obvious one

bobular:
Yeah sorry guys - I think I was the one to suggest equal width, but it doesn't look nice, I agree!

PR #148, opened 2021-06-09
adding radio button item's margin-right prop

moontrip:
added a new prop to control radio button item's margin-right prop for possibly better alignment of items. A screenshot is attached where itemMarginRight was set to 50

radiowidget2


jtlong3rd (approved):

bobular (approved):
Looks good DK, thanks.

I just saw the question about the mouseover `setFocused()` - it's probably an old question, but this is something that can be taken care of during the refactoring of plot controls.

bobular:
Removed Dave from reviewers because I'd like this merged so I can finalise the histogram controls (long story).

moontrip:
@bobular I will merge this. As you noticed I disabled setfocus as Steve didn't like it - it actually causes the component loading so perhaps we may not need it.

PR #150, opened 2021-06-14
Plot props/API tidy-up

bobular:
Ready for review.

[Developer documentation is here](https://github.com/VEuPathDB/web-components/blob/plot-props-tidy/docs/ADDING_A_NEW_PLOT.md)

We have accumulated a variety of different ways of handling plot sizing, and control of things like "displayLibraryControls", "staticPlot/interactive", spinner, margins and legend styling - a lot of these things are re-implemented in actual plot components (e.g. Histogram or Boxplot) but could and should easily be handled by the PlotlyPlot wrapper.  

Also after this refactor, no developer-facing Plot component props will have dependencies on Plotly.  (PlotlyPlot will still, of course.)

I have started to tidy-up and standardise our code.

The following has now been resolved, thanks to suggestions from Dave
> One thing I would like early feedback on is the `data` prop.
> 
> The new way of doing thing is that a component like Boxplot has a `boxplotData` prop (it used to be called `data`) - the Boxplot component processes `boxplotData` into `data` and passes it to the PlotlyPlot wrapper.
> 
> Is there a way to do it with the data props always called `data`?
> I guess it could always be called `plotData` but because it's defined in each module, it's OK to call it `boxplotData`, `histogramData`, etc.


bobular:
Note: I know Boxplot was a bad example to start with...
We already have a [commit](https://github.com/VEuPathDB/web-components/commit/fc032c0454e3c9605390563bcf80c62888cd9922) in the `boxplot` branch that removes the Plotly `Datum` dependency.  I haven't cherry-picked that commit into this repo yet.  I will work on the other plot components after getting feedback on the data prop.

src/plots/PlotlyPlot.tsx
```diff
@@ -18,41 +12,51 @@ type legendProp = {
...
  *
- * @param props
+ * @param props : PlotProps & PlotParams
+ *
+ * Takes all Plotly props (PlotParams) and our own PlotProps for
+ * controlling global things like spinner, library controls etc
+ *
  */
-export default function PlotlyPlot(props: PlotParams) {
+export default function PlotlyPlot(props: PlotProps & PlotParams) {
```
bobular, line 43: This is the big change I would like feedback on.

dmfalke:
@bobular, I don't understand your question about the `data` param.

dmfalke:
@bobular would this address your question?

```typescript
interface PlotProps<T> {
  data: T;
  ...
}

interface BoxPlotData {
  ...
}

interface BoxPlot extends PlotProps<BoxPlotData> {
  ...
}
```

bobular:
Thanks - it might! I hadn't considered that approach.  Will keep you posted.

dmfalke:
I think you will also need to do this:

```typescript
export default function PlotlyPlot(props: Exclude<PlotProps, 'data'> & PlotParams) {
```

bobular:
Aha, thanks, I did wonder about doing an Exclude to deal with this.

To avoid that I renamed the `data` prop on "our side" as `plotData`.  But
then there was no nice way to ensure that components extending PlotProps
contained plotData.  I think this does the trick, thanks!!

bobular:
Thanks again @dmfalke - it worked with `Omit` - now the API is much cleaner :-)

bobular:
Fixes #148

bobular:
Fixes https://github.com/VEuPathDB/web-eda/issues/148 but this probably won't be automagically linked...

src/plots/Boxplot.tsx
```diff
@@ -91,34 +87,18 @@ export default function Boxplot({
...
     [independentAxis]: {
       title: independentAxisLabel,
+      range: data.length ? undefined : [1, 5], // avoids x==0 line
+      showgrid: false,
+      tickfont: data.length ? {} : { color: 'transparent' },
     },
     showlegend: false,
   };
-  return (
-    <div style={{ position: 'relative', width: width, height: height }}>
```
chowington, line 102: @bobular The issue of the spinner not being centered is due to this `div` being removed. It wraps the plot and the spinner and is given the same width and height of the plot. The spinner centers itself within its container (seen in `Spinner.tsx`, so it centers inside this div and therefore centers over the plot as well.

chowington: You could probably just reintroduce this div in `PlotlyPlot`, I imagine

src/stories/plots/MosaicPlot.stories.tsx
```diff
@@ -1,76 +1,80 @@
...
-    data={[[]]}
-    independentValues={[]}
-    dependentValues={[]}
-    independentLabel={''}
-    dependentLabel={''}
-    showSpinner={true}
-  />
-);
+export const EmptyDataLoading = Template.bind({});
+EmptyData.args = {
```
chowington, line 77: Should be `EmptyDataLoading.args`. Causes this story to break. I've done this a couple times myself when copy/pasting!

chowington:
Great job on this Bob! Looks like a very flexible system 👍

dmfalke:
I've skimmed through the changes, and this is looking really great. I plan to take a more thorough look later. One thought that occurred to me is how these changes will relate to faceting. Basically I want to make sure we are optimizing for that use case and won't need to reimplement a bunch of stuff.

src/stories/api/githubDates.ts
```diff
@@ -14,12 +15,25 @@ type EventData = {
 export const getCreatedDateData = async (
```
jtlong3rd, line 15: Consider either:

1. Deleting the `async` keyword. (It's redundant since you're not using `await`.)

OR

2. Replacing `.then` / `.catch` with `await` / `try-catch`.

src/components/plotControls/HistogramControls.tsx
```diff
@@ -255,7 +256,7 @@ export default function HistogramControls({
           rowGap: '0.3125em',
         }}
       >
-        {onOpacityChange && (
+        {opacity != undefined && onOpacityChange && (
```
jtlong3rd, line 259: I'd recommend `!= null` instead of `!= undefined`. Our [standard ESLint config](https://www.npmjs.com/package/@veupathdb/eslint-config) enforces the [`smart` version of the `eqeqeq` rule](https://eslint.org/docs/rules/eqeqeq#smart).

src/components/plotControls/HistogramControls.tsx
```diff
@@ -266,7 +267,7 @@ export default function HistogramControls({
       <div
         style={{ display: 'flex', flexWrap: 'wrap', paddingTop: '0.3125em' }}
       >
-        {toggleDisplayLegend && (
+        {displayLegend != undefined && toggleDisplayLegend && (
```
jtlong3rd, line 270: Ditto.

src/components/plotControls/HistogramControls.tsx
```diff
@@ -280,7 +281,7 @@ export default function HistogramControls({
             containerStyles={{ paddingRight: '1.5625em' }}
           />
         )}
-        {toggleLibraryControls && (
+        {displayLibraryControls != undefined && toggleLibraryControls && (
```
jtlong3rd, line 284: Ditto.

jtlong3rd (approved):
This looks great! I had two minor style suggestions.

src/plots/PlotlyPlot.tsx
```diff
@@ -1,76 +1,140 @@
...
-export interface PlotProps {
-  width?: number;
-  height?: number;
-  margin?: Partial<Margin>;
-  staticPlot?: boolean;
-  showModebar?: boolean | 'hover';
-  // add legend prop for positioning
-  legend?: legendProp;
-  // show a loading spinner on top of the plot
+export interface PlotProps<T extends UnionOfPlotDataTypes> {
```
dmfalke, line 9: This part gives me pause. I don't think it is necessary to specify `T extends UnionOfPlotDataTypes`. I think `T` can simply be generic.

```suggestion
export interface PlotProps<T> {
```

dmfalke (changes requested):
So far, I only have one requested change.

src/plots/Boxplot.tsx
bobular: Thanks - I think I tried that and it didn't work...  But div placement is above my payscale ;-)

src/stories/api/githubDates.ts
bobular: Hmm, it compiles and works with or without the `async` keyword there.
But without it, my editor advises "This may be converted to an async function [80006]"
It's too late (at night) for me to learn more about this right now.

bobular:
> I've skimmed through the changes, and this is looking really great. I plan to take a more thorough look later. One thought that occurred to me is how these changes will relate to faceting. Basically I want to make sure we are optimizing for that use case and won't need to reimplement a bunch of stuff.

So far I've been hoping that standardising things like `displayLegend` (or is it `showLegend` lol) will be put us in good stead for faceting, but maybe there is more to it!

src/components/plotControls/HistogramControls.tsx
bobular: Thanks - I've found a few more elsewhere in the codebase.

Presumably `foo == undefined` should be `foo == null` too.
I'll do it more carefully in the morning though ;-)

src/stories/api/githubDates.ts
bobular: I did some more reading and added an `await` before the fetch.
Not sure why none of the previous variants caused compiler or IDE warnings!
Thanks for the pointers :+1: 
```
export const getCreatedDateData = async (
  url: string
): Promise<Array<EventData>> => {
  return await fetch(url)
    .then((response) => {
      if (response.ok) {
        return response.json();
      } else {
        throw new Error(
          'Response not as expected. Probably exceeded API rate limit.'
        );
      }
    })
    .then((json) => {
      return json.map((item: any) => ({
        id: item.id,
        date: item.created_at,
      }));
    })
    .catch((error) => {
      console.log(error.message);
    });
};
```

src/components/plotControls/HistogramControls.tsx
bobular: Cleaned up the whole repo in this regard (=== undefined -> == null,  !=undefined -> != null, etc) - tested affected components via checking stories.  All good

src/plots/PlotlyPlot.tsx
bobular: Thanks - I thought it would be helpful to constrain the XxxData types to ones we knew about, but I suppose it's a pain to maintain UnionOfPlotDataTypes.  The latter could probably be retired at some point (it is mainly a usePlotControls thing, I think) but not now.

src/plots/Boxplot.tsx
bobular: Tried again - worked this time. Thanks!

dmfalke (approved):

moontrip (approved):

chowington (approved):

src/stories/api/githubDates.ts
dmfalke: The primary purpose of using `async` is to avoid using Promise chains. While the code you have above is perfectly valid, it's not entirely conventional.

First, there isn't any real value in adding the `await` in the code above. It pretty much equivalent to the previous code.

A more conventional use of an `async` function would look like this (note the use of `await` to "unwrap" promises so that you can interact directly with the value.

```typescript
export const getCreatedDateData = async (
  url: string
): Promise<Array<EventData>> => {
  try {
    const response = await fetch(url);
    if (!response.ok) throw new Error(
      'Response not as expected. Probably exceeded API rate limit.'
    );
    const json = await response.json();
    return json.map((item: any) => ({
      id: item.id,
      date: item.created_at,
    }));
  }
  catch (error) {
    console.log(error.message);
  }
};
```

jtlong3rd: Just so we're all on the same page, @dmfalke's suggestion is exactly what I had in mind for `Replacing .then / .catch with await / try-catch`.

dmfalke (approved):

src/stories/api/githubDates.ts
bobular: Thanks both.  I have done as suggested. It definitely reads nicer!

PR #152, opened 2021-06-15
legend title example at XYPlot component

moontrip:
examined and tested to add legend title using XYPlot component. If acceptable, then I will add this approach to other plot components. A screenshot with an exemplary title of `legend title example` is attached:

legend-title-example

### DIFF
--- a/src/plots/XYPlot.tsx
+++ b/src/plots/XYPlot.tsx
@@ -7,7 +7,7 @@ import PlotlyPlot, { PlotProps } from './PlotlyPlot';
 import { Layout } from 'plotly.js';
 import Spinner from '../components/Spinner';
 // import extended legend prop
-import { layoutLegend } from '../types/general';
+import { LayoutLegendTitle } from '../types/plotly-omissions';
 
 export interface ScatterplotProps extends PlotProps {
   /** Data for the scatter plot */
@@ -76,7 +76,7 @@ export default function XYPlot(props: ScatterplotProps) {
     legendTitle,
   } = props;
 
-  const layout: Partial<Layout> & layoutLegend = {
+  const layout: Partial<Layout> & LayoutLegendTitle = {
     xaxis: {
       title: independentAxisLabel ? independentAxisLabel : '',
       range: independentAxisRange, // set this for better display: esp. for CI plot
--- a/src/plots/XYPlot.tsx
+++ b/src/plots/XYPlot.tsx
@@ -6,6 +6,8 @@ import PlotlyPlot, { PlotProps } from './PlotlyPlot';
 // import Layout for typing layout, especially with sliders
 import { Layout } from 'plotly.js';
 import Spinner from '../components/Spinner';
+// import extended legend prop
+import { layoutLegend } from '../types/general';
 
 export interface ScatterplotProps extends PlotProps {
   /** Data for the scatter plot */
@@ -59,6 +61,8 @@ export interface ScatterplotProps extends PlotProps {
   displayLegend?: boolean;
   /** show plotly's built-in controls */
   displayLibraryControls?: boolean;
+  /** legend title */
+  legendTitle?: string;
 }
 
 export default function XYPlot(props: ScatterplotProps) {
@@ -69,9 +73,10 @@ export default function XYPlot(props: ScatterplotProps) {
     independentAxisRange,
     dependentAxisRange,
     data,
+    legendTitle,
   } = props;
 
-  const layout: Partial<Layout> = {
+  const layout: Partial<Layout> & layoutLegend = {
     xaxis: {
       title: independentAxisLabel ? independentAxisLabel : '',
       range: independentAxisRange, // set this for better display: esp. for CI plot
@@ -93,6 +98,16 @@ export default function XYPlot(props: ScatterplotProps) {
     // plot title
     title: {
       text: title ? title : undefined,
+    },
+    // define legend.title
+    legend: {
+      title: {
+        text: legendTitle,
+        font: {
+          size: 14,
+          color: 'black',
+        },
+      },
     },
   };
 
--- a/src/types/general.ts
+++ b/src/types/general.ts
@@ -3,6 +3,9 @@
  * Or, at least, haven't found a more specific home yet.
  */
 
+// import Legend to extend
+import { Legend } from 'plotly.js';
+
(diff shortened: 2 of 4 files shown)


src/plots/XYPlot.tsx
```diff
@@ -69,9 +73,10 @@ export default function XYPlot(props: ScatterplotProps) {
     independentAxisRange,
     dependentAxisRange,
     data,
+    legendTitle,
   } = props;
 
-  const layout: Partial<Layout> = {
+  const layout: Partial<Layout> & layoutLegend = {
```
bobular, line 79: I think it would be good to stick with our unspoken rule of capitalising type names - and also rename it `LayoutLegendTitle`

src/types/general.ts
```diff
@@ -40,3 +43,19 @@ export type TimeDelta = { value: number; unit: string };
...
+      text?: string;
+      font?: {
+        family?: string;
+        size?: number;
+        color?: string;
+      };
+      side?: 'top' | 'left' | 'top left';
+    };
+  };
+}
```
bobular, line 61: I have two suggestions to tidy this up a bit, by removing all the question marks, and wrapping each inner level in `Partial<>` like they do in the index.d.ts of plotly.js

Then line 79 of XYPlot.tsx can be 
```
  const layout: Partial<Layout> & Partial<LayoutLegendTitle> = {
```

However, another solution is to define all the props as required, then use RecursivePartial defined [here on stackoverflow](https://stackoverflow.com/a/51365037/1909761) 
as follows:
```
  const layout: Partial<Layout> & RecursivePartial<LayoutLegendTitle> = {
```

src/types/general.ts
```diff
@@ -40,3 +43,19 @@ export type TimeDelta = { value: number; unit: string };
 export type NumberOrTimeDelta = number | TimeDelta;
 export type NumberOrDateRange = NumberRange | DateRange;
 export type NumberOrTimeDeltaRange = NumberRange | TimeDeltaRange;
+
+// extend legend type: somehow plotly.js did not define legend.title type
+// see https://plotly.com/javascript/reference/layout/#layout-legend-title
+export interface layoutLegend extends Partial<Legend> {
```
bobular, line 49: With my testing, the `extends Partial<Legend>` was not necessary.

bobular (changes requested):
Thanks DK. That's a good workaround to the annoying lapse in plotly TS.  Should we report a bug?

My minimal requested change is the renaming of the type with a capital letter. The other TS gurus may want to comment on the style suggestions regarding partial/optional props.

bobular:
Also, maybe move the extra type definition to a new *.ts file?   `src/types/plotly-omissions.ts` maybe?

Then if/when the problem is solved upstream, we can just remove the file, rather than risk (very small risk, I know) a merge conflict in general.ts.

src/types/general.ts
bobular: Reading more of https://stackoverflow.com/questions/41980195/recursive-partialt-in-typescript before I closed the tab... it seems like that is a risky path to take.

Probably best to keep the '?'s or use `Partial<>` like here in plotly's index.d.ts
```
export interface Layout {
    colorway: string[];
    title:
        | string
        | Partial<{
              text: string;
              font: Partial<Font>;
              xref: 'container' | 'paper';
              yref: 'container' | 'paper';
              x: number;
              y: number;
              xanchor: 'auto' | 'left' | 'center' | 'right';
              yanchor: 'auto' | 'top' | 'middle' | 'bottom';
              pad: Partial<Padding>;
          }>;
    titlefont: Partial<Font>;
    ...
```

moontrip:
@bobular thank you for your review and suggestions. As discussed in Slack, I will wait for others' comment and if these are good, then you may commit your local changes to this PR: certainly I can do that too if you want :)

bobular:
> @bobular thank you for your review and suggestions. As discussed in Slack, I will wait for others' comment and if these are good, then you may commit your local changes to this PR: certainly I can do that too if you want :)

Sorry I've already rolled them back - I was just playing around to see what would compile.  But I sent you the diff in case there's anything useful in there.

moontrip:
@bobular yes you already described them at our slack chat: anyway your explanations are clear so I can handle it anyway :) As you suggested in the end, will make another ts file to have the legend.title type and then use the original(?) one with ? for now.

moontrip:
@bobular I have updated this PR based on your suggestions such as changing the prop name to `LayoutLegendTitle`, placing its  type definition to the new file, `plotly-omissions.ts`, and removed `Legend` type import. As mentioned in my previous comment, I left the use of `?` for now following your comments.

jtlong3rd (approved):

bobular (approved):
Great.  I will probably also use plotly-omissions.ts in the general refactoring. It seems that type definitions are missing in several places.

moontrip:
Thank you for your reviews @jtlong3rd and @bobular 👍 

@bobular As I said in the last meeting, I will wait for your works (not merge) even if this ticket is approved :)

bobular:
> Thank you for your reviews @jtlong3rd and @bobular +1
> 
> @bobular As I said in the last meeting, I will wait for your works (not merge) even if this ticket is approved :)

Thanks for the consideration @moontrip 

I have just checked this PR against my proposed changes and they will not cause any conflicts, so please merge to `main` and then I can merge from here if needed.  Thanks!

moontrip:
@bobular thank you for your check for conflicts :) 
@dmfalke any review/comment ?

dmfalke:
@moontrip, sorry for the delayed review. I read through the comments, and I think this looks good. One thing I'd like to consider is if it is possible to use relative font sizes for the legend title (and possibly other text that plotly draws). It's not urgent, but in the future we might want/need to be able to scale a plot, etc.

moontrip:
@dmfalke Thank you for your review and comment 👍 Actually I tried to use relative font size like `em` in there, however, the `legend.title.font.size` parameter at plot does not accept string: only allows number. The legend title is actually handled as `SVG.text` in the plotly: if I remember correctly, `SVG.text` has font-size attribute, thus I think it can also be handled via CSS later.

bobular:
I agree. During the refactor tidy-up I'm removing as much
arbitrary/absolute styling as possible in order to pave the way for
something more flexible/maintainable.

PR #161, opened 2021-06-28
handling date for scatter plot etc

moontrip:
This PR involves several works concerning scatter plot and its viz (at web-eda)

Detailed description and some screenshots can be found at [the corresponding viz](https://github.com/VEuPathDB/web-eda/pull/190)

Note: these involve significant changes from previous scatter plot (XYPlot) component, thus merging with plot tidy-up work by Bob may need to be done carefully

### DIFF
--- a/src/components/plotControls/XYPlotControls.tsx
+++ b/src/components/plotControls/XYPlotControls.tsx
@@ -36,6 +36,7 @@ export type XYPlotControlsProps = {
   /** button color: for now, supporting blue and red only - primary: blue; secondary: red */
   buttonColor?: 'primary' | 'secondary';
   /** margin of radio button group: string array for top, left, bottom, and left, e.g., ['10em', '0', '0', '10em'] */
+  // TO DO: standardise to use SpacingOptionsAddon?
   margins?: string[];
   /** marginRight of radio button item: default 16px from MUI */
   itemMarginRight?: number | string;
--- a/src/components/plotControls/XYPlotControls.tsx
+++ b/src/components/plotControls/XYPlotControls.tsx
@@ -8,7 +8,6 @@ import { ErrorManagement } from '../../types/general';
 // Local Components
 import RadioButtonGroup from '../widgets/RadioButtonGroup';
 import Notification from '../widgets/Notification';
-import LabelledGroup from '../widgets/LabelledGroup';
 
 /**
  * Props for XYPlot controls.
@@ -26,7 +25,7 @@ export type XYPlotControlsProps = {
   errorManagement: ErrorManagement;
   /** XYPlot: valueSpec */
   valueSpec?: string;
-  /** XYPlot: valueSpec */
+  /** XYPlot: onValueSpecChange */
   onValueSpecChange?: (valueSpec: string) => void;
   /** How buttons are displayed. Vertical or Horizontal */
   orientation?: 'vertical' | 'horizontal';
@@ -40,6 +39,8 @@ export type XYPlotControlsProps = {
   margins?: string[];
   /** marginRight of radio button item: default 16px from MUI */
   itemMarginRight?: number | string;
+  /** plot options as props */
+  plotOptions?: string[];
 };
 
 /**
@@ -60,6 +61,8 @@ export default function XYPlotControls({
   buttonColor,
   margins,
   itemMarginRight,
+  // add plotOptions
+  plotOptions,
 }: XYPlotControlsProps) {
   const { ref, width } = useDimensions<HTMLDivElement>();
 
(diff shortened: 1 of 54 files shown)


src/components/plotControls/XYPlotControls.tsx
```diff
@@ -99,11 +102,8 @@ export default function XYPlotControls({
         {valueSpec && onValueSpecChange && (
           <RadioButtonGroup
             label="Plot options"
-            options={[
-              'Raw',
-              'Smoothed mean with raw',
-              'Best fit line with raw',
-            ]}
+            // following plotOptions
```
bobular, line 105: Good to see the EDA-specific options move to a prop.

I suggest not using XYPlotControls at all.

Just "roll your own" controls like I did in the [histogram viz](https://github.com/VEuPathDB/web-eda/blob/main/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx#L344)   (note: I forgot to remove the HistogramControls import!)

For example, the "Plot options" RadioButtonGroup is small/simple enough to use directly in ScatterplotVisualization.tsx - in my opinion.

If a control becomes more than a few lines, then create a new component like e.g. [BinWidthControl](https://github.com/VEuPathDB/web-components/blob/main/src/components/plotControls/BinWidthControl.tsx)

But don't delete anything!   Maybe move forward with the roll-your-own approach with Boxplot (and we can discuss it in our meeting)

src/plots/XYPlot.tsx
```diff
@@ -54,25 +54,31 @@ export interface ScatterplotProps extends PlotProps {
...
+  /** independentValueType 'number' (default) or 'date' (x data should be given as string[])  */
+  independentValueType?: 'number' | 'date';
+  /** dependentValueType 'number' (default) or 'date' (y data should be given as string[])  */
+  dependentValueType?: 'number' | 'date';
   /** legend title */
   legendTitle?: string;
 }
 
-export default function XYPlot(props: ScatterplotProps) {
+export default function XYPlot(props: XYPlotProps) {
```
bobular, line 72: Main task now (as I'm sure you're aware) is to merge your work here with what I did in the `plot-props-tidy` branch (now in `main`)

bobular (commented):
Thanks DK. I'll do a final review after you've resolved the merge conflicts with `main` (from `plot-props-tidy`)

src/components/plotControls/XYPlotControls.tsx
```diff
@@ -99,11 +102,8 @@ export default function XYPlotControls({
         {valueSpec && onValueSpecChange && (
           <RadioButtonGroup
             label="Plot options"
```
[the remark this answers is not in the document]
moontrip: @bobular actually I have a plan to use label as a prop from parent as well when merging your tidy-up :)

src/components/plotControls/XYPlotControls.tsx
moontrip: @bobular yes we have discussed the case to use RadioButtonGroup directly in our last meeting. I just waited for your works (tidy-up) to be merged, then would work for all things at once.

src/plots/XYPlot.tsx
moontrip: @bobular certainly yes :) I have a plan to merge your works to both scatter and plot components and their vizs. It would be tricky but perhaps manageable 🤞

moontrip:
Updated with tidy-up works in conjunction with the corresponding viz part. Details can be found at the [corresponding PR](https://github.com/VEuPathDB/web-eda/pull/190#issuecomment-877346889)

src/components/plotControls/XYPlotControls.tsx
```diff
@@ -100,11 +103,8 @@ export default function XYPlotControls({
...
-            options={[
-              'Raw',
-              'Smoothed mean with raw',
-              'Best fit line with raw',
-            ]}
+            // following plotOptions
+            options={plotOptions ? plotOptions : []}
```
jtlong3rd, line 107: Nitpick: `plotOptions ?? []`

jtlong3rd (commented):
Except for one tiny suggestion, looks good to me!

jtlong3rd (approved):

src/components/plotControls/XYPlotControls.tsx
moontrip: @jtlong3rd Thank you for your quick review and suggestion 👍 Actually we are trying to avoid XYPlotControls component so I did not carefully check it at this time but your are absolutely correct :) I have also made another small changes in the component and its story (again not that important part but since we are subject to the component...): making a new prop label so that it can be tested at story too.

moontrip:
I presume that Bob agreed with this PR as he approved corresponding viz PR :)

PR #162, opened 2021-07-01
Faceted plot implementation proposal

bobular:
Here is a `web-components`-only approach to implementing faceting that I would like feedback and suggestions for

Check out the last two stories in PiePlot.stories.tsx to see how faceted data looks (very simple version for now)

### DIFF
--- a/src/plots/FacetedPlot.tsx
+++ b/src/plots/FacetedPlot.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
+import { isFaceted } from '../../lib/types/guards';
 import { FacetedData } from '../types/plots';
 import { PlotProps } from './PlotlyPlot';
 
 export interface FacetedPlotProps<D, P extends PlotProps<D>> {
-  data: FacetedData<D>;
-  component: React.ElementType<P>; // ensures that the component type and props are compatible
+  data?: FacetedData<D> | D;
+  component: React.ElementType<P>;
   props: P;
 }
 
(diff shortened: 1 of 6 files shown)


bobular:
Do not be afraid of the failed checks.  Just go to the Pie plot stories to test this out.

I have [looked into plotless plots](https://github.com/VEuPathDB/web-components/issues/130) (with just the legend) and I didn't find anything promising.  We may have to implement our own.  This should be tied to the colour scheme/palette management.

bobular:
> Hey Bob, I'm just reviewing the code (not testing anything). This looks great. I'm not seeing a strong reason for `FacetedPlot` to accept non-faceted data. That would reduce some of the branching code. Thoughts?

I don't know. Don't we want to avoid code like this in the viz components? (Which adds a branch.)

```
  { isFaceted(data)
    ? (<FacetedPlot data={data} component={FooComponent} props={plotProps} />)
    : (<FooComponent data={data} {...plotProps})
  }
```
I guess it's OK if all the plotProps are bundled up into a variable as I've done above.

bobular:
I would like to merge this and the sibling components PR ASAP but this will require some basic wiring of ALL the visualizations the same as I already did for Barplot.  Could take an hour for each.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -24,15 +24,17 @@ A clear and concise description of what you expected to happen.
 If applicable, add screenshots to help explain your problem.
 
 **Desktop (please complete the following information):**
- - OS: [e.g. iOS]
- - Browser [e.g. chrome, safari]
- - Version [e.g. 22]
+
```
bobular, line 27: I didn't make these changes!

dmfalke: It looks like this branch is behind `main`. I'm sure if you merge the latest, this will go away.

dmfalke (approved):

jtlong3rd (approved):

PR #163, opened 2021-07-05
Add independentAxisRange prop to Histogram and custom validators to number/date inputs

bobular:
This is needed for https://github.com/VEuPathDB/web-eda/issues/198 because the histogram usually scales its x-axis to the min/max of the data.  Now you can specify a **wider** range with this prop, but not a narrower one.  This means that all data bins are always shown.  (If you need to zoom in, then provide zoomed data.)

An ugly story is added to both of the SelectedRange stories (number and date) to demo the independentAxisRange control.

### DIFF
--- a/src/components/widgets/NumberAndDateInputs.tsx
+++ b/src/components/widgets/NumberAndDateInputs.tsx
@@ -15,6 +15,12 @@ type BaseProps<M extends NumberOrDate> = {
   maxValue?: M;
   /** If true, warn about empty value. Default is false. */
   required?: boolean;
+  /** Optional validator function. Should return {validity: true, message: ''} if value is allowed.
+   * If provided, minValue and maxValue and required will have no effect.
+   */
+  validator?: (
+    newValue?: NumberOrDate
+  ) => { validity: boolean; message: string };
   /** Function to invoke when value changes. */
   onValueChange: (newValue?: NumberOrDate) => void;
   /** UI Label for the widget. Optional */
@@ -65,6 +71,7 @@ function BaseInput({
   value,
   minValue,
   maxValue,
+  validator,
   required = false,
   onValueChange,
   label,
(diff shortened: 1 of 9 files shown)


dmfalke (commented):
This looks good. I'm not sure I agree that we shouldn't allow for ranges that only include partial data. Can you explain the rationale? Thanks!

bobular:
> This looks good. I'm not sure I agree that we shouldn't allow for ranges that only include partial data. Can you explain the rationale? Thanks!

By ranges with partial data, do you mean `{min: undefined, max: 200}` ?

I agree we should discuss this, but is it part of this ticket or PR?

src/components/widgets/NumberAndDateInputs.tsx
```diff
@@ -92,36 +104,40 @@ function BaseInput({
   // Cancel pending onChange request when this component is unmounted.
   useEffect(() => debouncedOnChange.cancel, []);
 
+  const _validator =
+    validator ??
+    useCallback(
+      (newValue?: NumberOrDate): { validity: boolean; message: string } => {
+        if (newValue == null) {
+          return {
+            validity: !required,
```
bobular, line 113: Probably some FP magic can replace the `{ boolean, string }` return object here.

dmfalke:
> > This looks good. I'm not sure I agree that we shouldn't allow for ranges that only include partial data. Can you explain the rationale? Thanks!
> 
> By ranges with partial data, do you mean `{min: undefined, max: 200}` ?
> 
> I agree we should discuss this, but is it part of this ticket or PR?

Sorry for the delayed reply. I was referring to not allowing a narrower range. Is there a technical reason for that limitation?

bobular:
> 
> 
> > > This looks good. I'm not sure I agree that we shouldn't allow for ranges that only include partial data. Can you explain the rationale? Thanks!
> > 
> > 
> > By ranges with partial data, do you mean `{min: undefined, max: 200}` ?
> > I agree we should discuss this, but is it part of this ticket or PR?
> 
> Sorry for the delayed reply. I was referring to not allowing a narrower range. Is there a technical reason for that limitation?

Kind-of, yes.  Here's an example where the last bin gets chopped off.

image

bobular:
All done now - ready to review again please.

If you follow the last few commits, starting with https://github.com/VEuPathDB/web-components/pull/163/commits/4ef5740314c3e5d6ffc7bed3efd69557737886ad I removed the "range is never narrower than the binned data" logic before adding it back again because of the truncated bin in the screenshot above.

PR #165, opened 2021-07-15
only call .replace() on strings

dmfalke:
When the date type is `number`, a `ReferenceError` is thrown since `Number` doesn't have the `replace` method.


bobular (approved):
Nice widget!

PR #167, opened 2021-07-16
Histogram: add adjustBinEndToEndOfDay prop

bobular:
This allows us to handle two types of binEnd for dates - see https://github.com/VEuPathDB/EdaDataService/issues/54


bobular (commented):
I'm going to review my own PR here...

What this PR does is remove some default behaviour where binEnds were adjusted to the end of the binEnd's day (for date vars of course).

Now it is optional with `adjustBinEndToEndOfDay` and the default is NOT to do this.  This fits with the back end services now returning exclusive bin ranges (e.g. 2001-01-01 to 2001-02-01; previously the end date would have been 2001-01-31, e.g. inclusive) - and the EDA will not need to use this option.

PR #170, opened 2021-07-22
legend ellipsis with tooltip

moontrip:
Here is an implementation of legend ellipsis with tooltip which is a popular approach to deal with a long legend text. Note that this approach does not work at storybook but I could confirm that this just works fine at our vizs (tested at scatter and box plot vizs), so there may be some conflicts between plotly and storybook for this approach. So, if you want to test this, then please use `yarn link`

I think this approach may be useful for other purposes, e.g., changing plot beyond plotly's built-in function, etc.

Here are screenshots when mouse-over to the legends:

ellipsis-scatter1

ellipsis-box1

### DIFF
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -108,7 +108,7 @@ export default function Barplot({
       layout={layout}
       // stacked barplot reverses legend at plotly as plot start from bottom
       // default bar plot is grouped but just add this for a possible plot control in the future
-      reverseLegendTooltips={barLayout === 'stack' ? true : false}
+      reverseLegendTooltips={barLayout === 'stack'}
       {...restProps}
     />
   );
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -103,6 +103,13 @@ export default function Barplot({
   };
 
   return (
-    <PlotlyPlot data={plotlyFriendlyData} layout={layout} {...restProps} />
+    <PlotlyPlot
+      data={plotlyFriendlyData}
+      layout={layout}
+      // stacked barplot reverses legend at plotly as plot start from bottom
+      // default bar plot is grouped but just add this for a possible plot control in the future
+      reverseLegendTooltips={barLayout === 'stack' ? true : false}
+      {...restProps}
+    />
   );
 }
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -109,13 +109,6 @@ export default function Barplot({
   };
 
   return (
-    <PlotlyPlot
-      data={plotlyFriendlyData}
-      layout={layout}
-      // stacked barplot reverses legend at plotly as plot start from bottom
-      // default bar plot is grouped but just add this for a possible plot control in the future
-      reverseLegendTooltips={barLayout === 'stack'}
-      {...restProps}
-    />
+    <PlotlyPlot data={plotlyFriendlyData} layout={layout} {...restProps} />
   );
 }
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -7,6 +7,8 @@ import {
   OpacityDefault,
   OrientationAddon,
   OrientationDefault,
+  DependentAxisLogScaleAddon,
+  DependentAxisLogScaleDefault,
 } from '../types/plots';
 import PlotlyPlot, { PlotProps } from './PlotlyPlot';
 import { Layout } from 'plotly.js';
@@ -16,7 +18,8 @@ export interface BarplotProps
   extends PlotProps<BarplotData>,
     BarLayoutAddon<'overlay' | 'stack' | 'group'>,
     OrientationAddon,
-    OpacityAddon {
+    OpacityAddon,
+    DependentAxisLogScaleAddon {
   /** Label for independent axis. e.g. 'Country' */
   independentAxisLabel?: string;
   /** Label for dependent axis. Defaults to 'Count' */
@@ -42,6 +45,7 @@ export default function Barplot({
   barLayout = 'group',
   showIndependentAxisTickLabel = true,
   showDependentAxisTickLabel = true,
+  dependentAxisLogScale = DependentAxisLogScaleDefault,
   ...restProps
 }: BarplotProps) {
   // Transform `data` into a Plot.ly friendly format.
@@ -86,6 +90,8 @@ export default function Barplot({
 
   const dependentAxisLayout: Layout['yaxis'] | Layout['xaxis'] = {
     automargin: true,
+    tickformat: ',.1r', // comma-separated thousands, rounded to 1 significant digit
+    type: dependentAxisLogScale ? 'log' : 'linear',
     title: {
       text: dependentAxisLabel ? dependentAxisLabel : '',
     },
--- a/src/plots/Histogram.tsx
+++ b/src/plots/Histogram.tsx
@@ -378,7 +378,7 @@ export default function Histogram({
       onSelected={handleSelectedRange}
       onSelecting={handleSelectingRange}
       // stacked histogram (e.g., viz) reverses legend at plotly as plot start from bottom
-      reverseLegendTooltips={barLayout === 'stack' ? true : false}
+      reverseLegendTooltips={barLayout === 'stack'}
       {...restProps}
     />
   );
--- a/src/plots/Histogram.tsx
+++ b/src/plots/Histogram.tsx
@@ -10,6 +10,8 @@ import {
   OrientationAddon,
   OrientationDefault,
   BarLayoutAddon,
+  DependentAxisLogScaleAddon,
+  DependentAxisLogScaleDefault,
 } from '../types/plots';
 import { NumberOrDate, NumberOrDateRange, NumberRange } from '../types/general';
 
@@ -34,7 +36,8 @@ export interface HistogramProps
   extends PlotProps<HistogramData>,
     OrientationAddon,
     OpacityAddon,
-    BarLayoutAddon<'overlay' | 'stack'> {
+    BarLayoutAddon<'overlay' | 'stack'>,
+    DependentAxisLogScaleAddon {
   /** Label for independent axis. Defaults to `Bins`. */
   independentAxisLabel?: string;
   /** Label for dependent axis. Defaults to `Count`. */
@@ -42,8 +45,6 @@ export interface HistogramProps
   /** Range for the dependent axis (usually y-axis) */
   // can only be numeric
   dependentAxisRange?: NumberRange;
-  /** Use a log scale for dependent axis. Default is false */
-  dependentAxisLogScale?: boolean;
   /** Show value for each bar */
   showValues?: boolean;
   /** A range to highlight by means of opacity */
@@ -57,6 +58,8 @@ export interface HistogramProps
   isZoomed?: boolean;
   /** independent axis range min and max (this will be widened to include data if needed) */
   independentAxisRange?: NumberOrDateRange;
+  /** if true (default false), adjust binEnds to the end of the day */
+  adjustBinEndToEndOfDay?: boolean;
 }
 
 /** A Plot.ly based histogram component. */
@@ -68,13 +71,14 @@ export default function Histogram({
   opacity = OpacityDefault,
   barLayout = 'overlay',
   dependentAxisRange,
-  dependentAxisLogScale = false,
+  dependentAxisLogScale = DependentAxisLogScaleDefault,
   showValues,
   selectedRange,
   onSelectedRangeChange = () => {},
   selectedRangeBounds,
   isZoomed = false,
   independentAxisRange,
+  adjustBinEndToEndOfDay = false,
   ...restProps
 }: HistogramProps) {
   /**
@@ -123,7 +127,9 @@ export default function Histogram({
             return (
               DateMath.diff(
                 new Date(bin.binStart as string),
-                DateMath.endOf(new Date(bin.binEnd as string), 'day'),
+                adjustBinEndToEndOfDay
+                  ? DateMath.endOf(new Date(bin.binEnd as string), 'day')
+                  : new Date(bin.binEnd as string),
                 'seconds',
                 false
               ) * 1000
@@ -160,7 +166,14 @@ export default function Histogram({
           },
         };
       }),
-    [data, orientation, calculatedBarOpacity, selectedRange, showValues]
+    [
+      data,
+      orientation,
+      calculatedBarOpacity,
+      selectedRange,
+      showValues,
+      adjustBinEndToEndOfDay,
+    ]
   );
 
   /**
@@ -198,7 +211,9 @@ export default function Histogram({
               new Date(bin.binStart as string),
               DateMath.diff(
                 new Date(bin.binStart as string),
-                DateMath.endOf(new Date(bin.binEnd as string), 'day'),
+                adjustBinEndToEndOfDay
+                  ? DateMath.endOf(new Date(bin.binEnd as string), 'day')
+                  : new Date(bin.binEnd as string),
                 'seconds',
                 false
               ) * 500,
@@ -206,7 +221,13 @@ export default function Histogram({
             ).toISOString()
           : ((bin.binStart as number) + (bin.binEnd as number)) / 2.0,
     }));
-  }, [data.series, data.valueType, isZoomed, selectedRangeBounds]);
+  }, [
+    data.series,
+    data.valueType,
+    isZoomed,
+    selectedRangeBounds,
+    adjustBinEndToEndOfDay,
+  ]);
 
   // local state for range **while selecting** graphically
   const [selectingRange, setSelectingRange] = useState<NumberOrDateRange>();
@@ -270,7 +291,9 @@ export default function Histogram({
       const rightCoordinate =
         data.valueType === 'number'
           ? range.max
-          : DateMath.endOf(new Date(range.max), 'day').toISOString();
+          : adjustBinEndToEndOfDay
+          ? DateMath.endOf(new Date(range.max), 'day').toISOString()
+          : range.max;
       return [
         {
           type: 'rect',
@@ -302,7 +325,13 @@ export default function Histogram({
     } else {
       return [];
     }
-  }, [selectingRange, selectedRange, orientation, data.series]);
+  }, [
+    selectingRange,
+    selectedRange,
+    orientation,
+    data.series,
+    adjustBinEndToEndOfDay,
+  ]);
 
   const plotlyIndependentAxisRange = useMemo(() => {
     // here we ensure that no data bins are excluded/hidden from view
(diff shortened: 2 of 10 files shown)


moontrip:
Some updates are made following Bob and Dave's comments in the Slack.

src/plots/PlotlyPlot.tsx
```diff
@@ -54,9 +56,39 @@ export default function PlotlyPlot<T>(
...
+   * So, perhaps we need a new prop to know which plot component is calling this PlotlyPlot
+   * e.g., plotComponent={'Mosaic'} at <Mosaic>, and define plotComponent?: string at PlotProps
+   * Then, if(plotComponent === 'Mosaic') storedLegendList.reverse();
+   */
+
+  // set the number of characters to be displayed
+  const maxLegendText = 20;
+  // change data.name with ellipsis
+  for (let i = 0; i < data.length; i++) {
+    data[i].name =
```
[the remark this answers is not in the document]
bobular: perhaps it should be another useMemo transformation, something like
```
  const finalData = useMemo(() => data.map((d) => { ...d, name: substring_stuff }, [ data, maxLegendText ])
```

bobular (commented):
Added a few thoughts and suggestions.

I think we should wait until a decision is made on legends before doing any further work in this branch, however.

src/plots/PlotlyPlot.tsx
```diff
@@ -54,9 +56,39 @@ export default function PlotlyPlot<T>(
     legendTitle,
     spacingOptions,
     showSpinner,
+    // expose data for applying legend ellipsis
+    data,
     ...plotlyProps
   } = props;
 
+  // store legend list for tooltip
+  let storeLegendList: NonNullable<string[]> = [];
```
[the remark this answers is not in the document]
bobular: Also, I'd recommend a more noun-like name, such as `storedLegendList`

moontrip:
> Added a few thoughts and suggestions.
> 
> I think we should wait until a decision is made on legends before doing any further work in this branch, however.

Thank you for your review and comments 👍 I agree with your opinion that we need to wait for the decision which approach we would take. Anyway, I will leave my response to your comment individually

src/plots/PlotlyPlot.tsx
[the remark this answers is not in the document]
moontrip: Yes useMemo would work here but I was not quite sure whether there is a case to call plotlyplot component without changing data.

src/plots/PlotlyPlot.tsx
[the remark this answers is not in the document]
moontrip: yeah in general React preserves the rule, immutable, and I remember that the finalData was used in the previous plotlyplot component before your tidy-up.

src/plots/PlotlyPlot.tsx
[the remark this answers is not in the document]
moontrip: I thought of this case a bit more whether we should use useMemo or not. I guess that the number of data array is not that large, probably around 10 in max in most cases, which means about 10 operations/loops would be demanded to construct storeLegendList. Based on some articles, useMemo is not that efficient when less than 100 operations (or arguably even 1000 operations) is done because of its overhead, especially initial render. Thus, I think we can live without useMemo in this case 😉

src/plots/PlotlyPlot.tsx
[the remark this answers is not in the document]
moontrip: In addition to my thought earlier regarding useMemo, what if data is very large (e.g., millions points) ? It may create additional overhead as well.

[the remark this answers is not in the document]
bobular: Yeah I agree for scatterplot or boxplot with lots of outliers/rawData there might be a performance hit with useMemo() so we can try without it.

moontrip:
I have mostly addressed what Bob suggested, like change variable name, add new props, etc. It took more time than I expected because I found a couple of issues so addressed them as well. They are:

a) placing `storedLegendList` variable inside `PlotlyPlot` function produced inconsistent behavior for plot control of scatter plot viz at least. After testing several approaches, it turned out that putting it as global variable works just fine. Perhaps related to lazy loading ?
b) Mosaic reverses data so that the first data is shown at top. Added a new prop to check reversed data or not and then performed a reverse for  `storedLegendList` array.
c) Even after addressing Mosaic's reverse data, Mosaic plot did not work well with legend tooltip. I realized that it defines additional axis at data with empty values (perhaps for display purpose), thus needed to filter out the dataset to make legend list for tooltip.
d) Realized that in some cases, legend text with HTML tag (e.g., <br>, etc.) is broken if it is somehow located around the max number of truncated characters. This is the case for scatter plot where I have used a couple of HTML tags for legend text (e.g., square). Accordingly, it was addressed at this PR, https://github.com/VEuPathDB/web-eda/pull/258.

jtlong3rd (commented):
The general approach of passing an `onAfterPlot` to alter the legend looks sound!

Out of curiosity, what did your code for deriving `storedLegendList` within the `PlotyPlot` look like? It's suspicious that it didn't work - maybe `storedLegendList`, `onAfterPlot`, and/or `finalData` needs to be memoized? (Not for performance reasons, but to keep the references to `storedLegendList`, `onAfterPlot`, and/or `finalData` stable between renders.)

moontrip:
@jtlong3rd thank you for your review and comment 👍 At one point, I did define `storedLegendList` within the `PlotlyPlot` component (local variable) and pass it directly to `legendTooltip` function via `onAfterPlot` prop (see the link in the end). It worked fine in general except for the case when I used plot control of scatter plot viz (Raw, Smoothed mean with raw, and best fit line with raw radio buttons): realized this recently, not at initial implementation. With the same code shown in the link, the legend ellipsis for plot control behavior worked again by simply define `storedLegendList` as a global variable (out of `PlotlyPlot`). Honestly, I didn't test whether useMemo() for the local `storedLegendList` would resolve the issue as that simple move (local to global) resolved the issue.

https://github.com/VEuPathDB/web-components/commit/13b17e910dfa820755870d25c52fc51b539c75e8#

bobular:
I got it to work with

```
  const storedLegendList = useMemo(() => {
    if (data != null) {
      return data
	.filter((data) => {
          if (data.name == null) return false;
          return true;
	})
	.map((data) => {
          return data.name ?? '';
	});
    } else {
      return [];
    }
  }, [ data ]);
```
and
```
  const legendTooltip = useCallback(() => {
    select('g.legend')
      .selectAll('g.traces')
      .append('svg:title')
      .text((d, i) => storedLegendList[i]);
  }, [ storedLegendList ]);
```

I commented out the `reverse()` bit to keep things simple while testing this.

bobular:
Instead of the `reverse()` and the prop to enable it,  see if `d.trace.index` can be used instead of map's `i`

jtlong3rd:
> I got it to work with
> 
> ```
>   const storedLegendList = useMemo(() => {
>     if (data != null) {
>       return data
> 	.filter((data) => {
>           if (data.name == null) return false;
>           return true;
> 	})
> 	.map((data) => {
>           return data.name ?? '';
> 	});
>     } else {
>       return [];
>     }
>   }, [ data ]);
> ```
> 
> and
> 
> ```
>   const legendTooltip = useCallback(() => {
>     select('g.legend')
>       .selectAll('g.traces')
>       .append('svg:title')
>       .text((d, i) => storedLegendList[i]);
>   }, [ storedLegendList ]);
> ```
> 
> I commented out the `reverse()` bit to keep things simple while testing this.

Thanks for checking this out, @bobular!

@moontrip - do you think you could test this approach and see if that fixes the "inconsistent behavior for plot control of scatter plot viz" that you were observing before?

(Also, sorry I'm so fixated on this issue! It would be ideal to ditch the global variable, so as to avoid risking bugs around rendering multiple plots simultaneously.)

moontrip:
> > I got it to work with
> > ```
> >   const storedLegendList = useMemo(() => {
> >     if (data != null) {
> >       return data
> > 	.filter((data) => {
> >           if (data.name == null) return false;
> >           return true;
> > 	})
> > 	.map((data) => {
> >           return data.name ?? '';
> > 	});
> >     } else {
> >       return [];
> >     }
> >   }, [ data ]);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```
> >   const legendTooltip = useCallback(() => {
> >     select('g.legend')
> >       .selectAll('g.traces')
> >       .append('svg:title')
> >       .text((d, i) => storedLegendList[i]);
> >   }, [ storedLegendList ]);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I commented out the `reverse()` bit to keep things simple while testing this.
> 
> Thanks for checking this out, @bobular!
> 
> @moontrip - do you think you could test this approach and see if that fixes the "inconsistent behavior for plot control of scatter plot viz" that you were observing before?
> 
> (Also, sorry I'm so fixated on this issue! It would be ideal to ditch the global variable, so as to avoid risking bugs around rendering multiple plots simultaneously.)

First of all, thank you for suggesting hook approach, @bobular and your input @jtlong3rd 👍. I have tested the approach by putting `legendTooltip` callback into PlotlyPlot function due to the presence of the second argument in the useCallback hook, `[storedLegendList]`. Unfortunately, it did not work for me to resolve the issue on scatter plot viz's control. I wonder Bob tested the issue (switching plot mode via radio button)?

moontrip:
@jtlong3rd @bobular Since Bob's nice suggestions, I have been working for avoiding global variable of `storedLegendList` that also works for scatter plot viz. Trying a couple of possible approaches. Also found a couple of issues/bugs for specific cases that should be addressed. Will let you know about the result (🤞) after done.

bobular:
Hi guys - I've just come back to this - sorry I could only give it half an hour earlier.

My suggested approach actually was very unreliable. It seems to work directly after the code is recompiled, but not after changing between visualizations or between scatterplot modes. :-(

moontrip:
> Hi guys - I've just come back to this - sorry I could only give it half an hour earlier.
> 
> My suggested approach actually was very unreliable. It seems to work directly after the code is recompiled, but not after changing between visualizations or between scatterplot modes. :-(

@bobular No problem~ It was helpful to revisit this topic. Fortunately I found a solution and am preparing for push 😃

bobular:
That's great.  I felt guilty and looked at it a bit more.  Seems like an empty `<title></title>` is getting appended first, sometimes. And then a bunch of others ;-)

```
<title></title><title>Less frequent than a few times per week</title><title>Less frequent than a few times per week</title><title>Less frequent than a few times per week</title><title>Less frequent than a few times per week</title><title>Less frequent than a few times per week</title><title>Less frequent than a few times per week</title>
```

This may help with avoiding the reversing for some plot types (though I haven't tested with Mosaic)
```
.text((d) => storedLegendList[d[0].trace.index]);
```
(In a comment above I erroneously said it was `d.trace.index`)

moontrip:
@jtlong3rd (and @bobular ),

I found a solution to avoid global variable. Surprisingly, `onUpdate` event handler at built-in `Plot` component works like `onAfterPlot` that corresponds to `plotly_afterplot` in the pure plotly.js and is generally utilized when changing legend at pure plotly.js. So, for addressing the issue at scatter plot viz, I could use `onUpdate` instead of `onAfterPlot`, which works for all cases I have tested (with scatter (with control), bar, mosaic, histogram vizs, and subset's histogram (histogramfilter component) ). Bob's suggestions are mostly used in this update with slight modifications. 

There are a couple of more bug fixes for specific cases as follows:
a) some scatter plot data has empty data.value for smoothed mean case, while data.name is defined. In this case, corresponding trace and legend are not shown in the plot but data.name (legend text) still exists: i.e., unnecessary legend list is added to `storedLegendList`. Thus, it was necessary to filter them out at `storedLegendList`.
b) it seems like stacked plot at plotly places the data from the bottom (i.e., last data is on top) and it is the same to legend, while the order of data is not changed. For example, histogram viz utilizes `barMode: 'stack'`, which makes legend tooltip be reversed in order. Thus, I have added reverseLegendTooltips prop at histogram and barplot (just in case that we may introduce a control in there) and checked 'stack' with conditional operator. 
c) @jtlong3rd I have added useRef (div ref) in this update. Please let me know if it is what you thought: seemingly it worked fine for me from several tests.

src/plots/Barplot.tsx
```diff
@@ -103,6 +103,13 @@ export default function Barplot({
   };
 
   return (
-    <PlotlyPlot data={plotlyFriendlyData} layout={layout} {...restProps} />
+    <PlotlyPlot
+      data={plotlyFriendlyData}
+      layout={layout}
+      // stacked barplot reverses legend at plotly as plot start from bottom
+      // default bar plot is grouped but just add this for a possible plot control in the future
+      reverseLegendTooltips={barLayout === 'stack' ? true : false}
```
jtlong3rd, line 111: I would just pass `barLayout === 'stack'` to `reverseLegendTooltips`.

src/plots/Histogram.tsx
```diff
@@ -377,6 +377,8 @@ export default function Histogram({
       data={plotlyFriendlyData}
       onSelected={handleSelectedRange}
       onSelecting={handleSelectingRange}
+      // stacked histogram (e.g., viz) reverses legend at plotly as plot start from bottom
+      reverseLegendTooltips={barLayout === 'stack' ? true : false}
```
jtlong3rd, line 381: Ditto!

src/plots/PlotlyPlot.tsx
```diff
@@ -115,14 +134,70 @@ export default function PlotlyPlot<T>(
...
+
+  // reverse storedLegendList when reverseLegendTooltips is true (e.g., Mosaic's reversed data)
+  // default behavior of stacked plot at plotly is to put the first data and legend at the bottom (histogram, bar plot)
+  if (reverseLegendTooltips) storedLegendList.reverse();
+
+  // define useRef
+  const plotlyPlotRef = useRef<HTMLDivElement>(null);
+
+  // add legend tooltip function
+  const legendTooltip = useCallback(() => {
```
jtlong3rd, line 163: Since `onUpdate` receives the div into which the figure was rendered, we may not actually need a ref! Maybe something like this will work:

```typescript
  const onUpdate = useCallback((figure: Readonly<Figure>, graphDiv: Readonly<HTMLElement>) => {
    select(graphDiv)
      .select('g.legend')
      .selectAll('g.traces')
      .append('svg:title')
      .text((d, i) => storedLegendList[i]);
  }, [storedLegendList]);
```

jtlong3rd (commented):
Just a few small suggestions, and then I think this is good to merge!

bobular:
Nice one DK.  I'll do some thorough functional testing in the morning.

src/plots/Barplot.tsx
moontrip: I was overthinking 😄

src/plots/Histogram.tsx
moontrip: Ditto!

src/plots/PlotlyPlot.tsx
moontrip: Thank you for your suggestion! I wonder what the `figure` argument actually means here as it is not used in the function. Also, from my understanding,  `graphDiv` refers to a div id that need to be set at `<div style={{ ...containerStyles, position: 'relative' }}>`? e.g., 

```
...
const graphDiv = document.getElementById('graphDivId');
const onUpdate = useCallback...
...
<div id='graphDivId' style={{ ...containerStyles, position: 'relative' }}>
...
```

jtlong3rd: Yeah, I would probably name the `figure` argument `_` to signify that we're not using it. I played around with `onUpdate `a little and it looks like `graphDiv` is actually a reference an internal `div` node which Plotly draws. That's really convenient because you won't need to set some sort of `id` prop or set up a React `ref` to a containing `div`.

moontrip: Thank you for your detailed information! 👍 It is quite interesting as I didn't know that the graphDiv could be used like that without getting id. Yes it worked just fine with your approach - pushed those updates :)

jtlong3rd (approved):
Nice going, @moontrip!

bobular (commented):
Functionally seems solid now, but I have a few suggestions.

`maxLegendTextLength` prop could go in `PlotLegendAddon`

I'm also seeing multiple `<title>` svg elements added by `onUpdate` - more get added if toggling between log/no-log scale in Barplot or Histogram.  Maybe @jtlong3rd can help here?

image

moontrip:
> Functionally seems solid now, but I have a few suggestions in addition to the [legend indexing change](https://github.com/VEuPathDB/web-components/pull/170/commits/e8fd6ac4e283d8e2c2132aa0e3bda6e1c20113ac) I pushed to the branch.
> 
> `maxLegendTextLength` prop could go in `PlotLegendAddon`
> 
> I'm also seeing multiple `<title>` svg elements added by `onUpdate` - more get added if toggling between log/no-log scale in Barplot or Histogram. Maybe @jtlong3rd can help here?
> 
> image

Thank you for your update on the logic 👍 I have tested several cases that I usually did, however, unfortunately, it does not work for the case where data.value is empty. See the screenshot below that is GEMS1 data. As you can see, there are no smoothed mean (s-mean) and confidence interval (CI) legends for India and Kenya because data were not sufficient to generate s-mean and CI so corresponding data.value are empty while data.name still exists. Since there may be more unknown exceptions, I tried to handle it on data level like I used filter function.

ellipsis-error1

moontrip:
@bobular I don't know if repeated <title></title> is specifically related onUpdate as I didn't check it under onAfterPlot. Actually it also happens when you click legend text to show/hide trace regardless of onUpdate/onAfterPlot so I just leave it as is. 

If you don't like the behavior, perhaps one may check the presence of <title> at legend, remove them, and add <title> again in the process of onUpdate function (legend tooltip function). I don't know if it costs more than having multiple <title>.

bobular:
> @bobular I don't know if repeated <title></title> is specifically related onUpdate as I didn't check it under onAfterPlot. Actually it also happens when you click legend text to show/hide trace regardless of onUpdate/onAfterPlot so I just leave it as is.

It was happening before with onAfterPlot too :-)

moontrip:
Bob found a quick solution by removing `.filter()` so applied it 👍

moontrip:
@jtlong3rd @bobular I have tested the approach I said, removing title and regenerating it again. It works fine for the purpose to avoid duplicate titles, like (d3 part inside onUpdate function).  

```
      // remove pre-existing title to avoid duplicates
      select(graphDiv)
        .select('g.legend')
        .selectAll('g.traces')
        .selectAll('title').remove()
      // add tooltip
      select(graphDiv)
        .select('g.legend')
        .selectAll('g.traces')
        .append('svg:title')
        .text((d) => storedLegendList[d[0].trace.index]);
```

jtlong3rd:
> @jtlong3rd @bobular I have tested the approach I said, removing title and regenerating it again. It works fine for the purpose to avoid duplicate titles, like (d3 part inside onUpdate function).
> 
> ```
>       // remove pre-existing title to avoid duplicates
>       select(graphDiv)
>         .select('g.legend')
>         .selectAll('g.traces')
>         .selectAll('title').remove()
>       // add tooltip
>       select(graphDiv)
>         .select('g.legend')
>         .selectAll('g.traces')
>         .append('svg:title')
>         .text((d) => storedLegendList[d[0].trace.index]);
> ```

I think this approach is fine- let's go with it!

(The only other possibility that came to mind was trying `onInitialize` instead of `onUpdate`, but I think `onInitialize` won't work because I don't think `onInitialize` is called every time `data` changes.)

jtlong3rd (approved):

PR #171, opened 2021-07-26
CategoricalAxis addOn for axis tick ordering

bobular:
This is **almost** what we need for ordering plots with categorical independent axes.

You can play around with the new props in `categoricalAxisOrderMethod` and `categoricalAxisCategoryOrder`
http://localhost:62331/?path=/story/plots-barplot--basic

If you add a category (elephants) without data to the order prop `["cats","elephants","dogs","monkeys"]` then it shows up as you would want on the plot
image

UNLESS it is the first or last element in the list:
`["elephants","cats","dogs","monkeys"]`
`["cats","dogs","monkeys","elephants"]`

So perhaps this is a plotly bug we can report and fix?

### DIFF
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -3,6 +3,8 @@ import { PlotParams } from 'react-plotly.js';
 import {
   BarLayoutAddon,
   BarplotData,
+  CategoricalAxisAddon,
+  categoricalAxisPlotlyProps,
   OpacityAddon,
   OpacityDefault,
   OrientationAddon,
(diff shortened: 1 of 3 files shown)


dmfalke (commented):
It seems like we should always display labels in the order that we receive the data. Is there a good reason not to?

See related comments [here](https://github.com/VEuPathDB/web-eda/pull/254#pullrequestreview-716927276).

bobular:
Ordering our heterogenous data is non-trivial but of course doable (and a challenge I embrace).

Fixing Plotly to handle the missing elephants (see above) seems harder - after I spent a good 20 mins trying to figure out where in the code they do it (and failing).

PR #175, opened 2021-08-03
Excessive rounding on y-axis tick numbers bugfix

bobular:
Spotted by @moontrip in https://github.com/VEuPathDB/web-eda/pull/254#issuecomment-891205601 

Actually we're not out of the woods yet.

As Dave points out, the mouse-over numbers are also affected by the rounding (now in logscale mode only).  This needs to be addressed through further digging into the Plotly manual.

### DIFF
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -90,7 +90,7 @@ export default function Barplot({
 
   const dependentAxisLayout: Layout['yaxis'] | Layout['xaxis'] = {
     automargin: true,
-    tickformat: ',.1r', // comma-separated thousands, rounded to 1 significant digit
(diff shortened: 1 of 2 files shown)


moontrip:
@bobular It works fine to me 👍

bobular:
Oh sorry @moontrip - I didn't realise I could edit other people's comments!!

bobular:
When in logscale mode, I had to set plotly's `yaxis.hoverformat` to something sensible.
`undefined` doesn't work, because then it falls back to `yaxis.tickformat`.
There doesn't seem to be a way to tell it to use "default formatting" so had to tell it to format zero decimals for integer count data and .4f for fractional/proportional data.  I detect which type of data it is, from the data.

dmfalke (approved):
This looks good.

Issue #151, opened 2021-06-14
Visualisations: Mosaic plot updates to axis labels


dmfalke:
1. Remove “(%)” from the end of both the x-axis and y-axis labels
2. Remove the proportion numbers from the x-axis, and replace with the values for the x-axis variable
3. Put “Proportion” in front of (or on top of) the y-axis label
4. Add the y-axis variable label to the legend
5. _(Possibly remove the counts and proportions next to the x-axis values)_

![image](https://user-images.githubusercontent.com/365139/121952463-40896a00-cd2a-11eb-8d6b-ca3431a0f70e.png)

bobular:
Please see #150 - merge in first before starting this!

aaronwlsong:
QA checking:

**QA 1. For the same study and the same value of x-axis and y-axis:** 

_1.Remove “(%)” from the end of both the x-axis and y-axis labels_
Not change on QA website, the (%)is still after y-axis label;

_2. Remove the proportion numbers from the x-axis, and replace with the values for the x-axis variable_
Not change on QA website, the proportion number still shows on x-axis;

_3. Put “Proportion” in front of (or on top of) the y-axis label_
Not change on website, there is no proportion on top /in front of the y-axis label;

_4.Add the y-axis variable label to the legend_
Not change on website, there is no legend;

_5. (Possibly remove the counts and proportions next to the x-axis values)_
Not found the counts next to the x-axis values

<img width="1546" alt="WeChat4414ef365a11056b0c63fddb034491c9" src="https://user-images.githubusercontent.com/61077845/130079489-1e7455ef-0a41-4142-9148-30af96b3f095.png">

**QA 2. For the same study and different value of y-axis:** 

_1.Remove “(%)” from the end of both the x-axis and y-axis labels_
Changed on QA website, the (proportion) is after y-axis label in stead of (%), but not for x-axis;

_2. Remove the proportion numbers from the x-axis, and replace with the values for the x-axis variable_
Changed on QA website, the value of x-axis shows on x-axis;

_3. Put “Proportion” in front of (or on top of) the y-axis label_
Change on QA website, there is (proportion) after the y-axis label;

_4.Add the y-axis variable label to the legend_
Changed on QA website, there is a legend;

_5. (Possibly remove the counts and proportions next to the x-axis values)_
Not changed.

[...]

PR #178, opened 2021-08-05
Mosaic minor changes

chowington:
Fixes https://github.com/VEuPathDB/web-components/issues/151

image

Note the "proportion" text on the left--slightly different than the spec, so I'll take feedback on it. Also, I opted to leave the x-axis labels counts and percentages in place.

### DIFF
--- a/src/plots/MosaicPlot.tsx
+++ b/src/plots/MosaicPlot.tsx
@@ -42,17 +42,7 @@ export default function MosaicPlot({
   });
 
   const layout = {
-    // Bottom x axis displaying percent ticks
     xaxis: {
-      title: independentAxisLabel,
-      // Must expliticly define range for it to work consistently
-      range: [0, 100] as number[],
-      tickvals: [0, 20, 40, 60, 80, 100] as number[],
-      ticktext: ['', '20', '40', '60', '80', '100'] as string[],
-      showgrid: true,
-    },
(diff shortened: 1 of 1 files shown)


chowington:
Actually since we're changing the wording to "proportion" rather than "%", should the y axis go from 0 to 1 instead of 0 to 100?

jtlong3rd:
> Actually since we're changing the wording to "proportion" rather than "%", should the y axis go from 0 to 1 instead of 0 to 100?

Worth asking a wider audience! I'm personally all for having the y-axis go from 0 to 1.

jtlong3rd (approved):

bobular (approved):
You could keep the percentages and just change the axis label suffix to "(Proportion (%))" maybe?

I tested in the EDA context.  This real-world example has not-very readable x-axis labels - but that would be OK if the mouseovers also showed the x-axis value.  Could `hovertext` at MosaicPlot.tsx:L75 include the x-value too?  Hmm, no I can see that that won't work, unless maybe we change hovermode to 'closest'.

PRISM2 border cohort
image

chowington:
I'll leave the 0-1 for later as it's not quite trivial and not a huge deal, and just go ahead and get this merged.
We certainly need to find a solution for the overlapping x-axis values.

PR #186, opened 2021-08-10
truncated X-axis

moontrip:
a draft PR as it only works for X-axis due to a couple of issues I have faced.

### Related issues
- https://github.com/VEuPathDB/web-eda/issues/98 (closes)

### DIFF
--- a/src/components/plotControls/AxisRangeControl.tsx
+++ b/src/components/plotControls/AxisRangeControl.tsx
@@ -20,6 +20,7 @@ export interface AxisRangeControlProps
   onRangeChange?: (newRange?: NumberOrDateRange) => void;
   /** DKDK default/initial independent axis range */
   defaultIndependentAxisRange?: NumberOrDateRange;
+  isAxisTruncated?: boolean;
 }
 
 export default function AxisRangeControl({
@@ -28,8 +29,9 @@ export default function AxisRangeControl({
   range,
   onRangeChange,
   containerStyles,
-  //DKDK
+  //DKDK pass defaultIndependentAxisRange
   defaultIndependentAxisRange,
+  isAxisTruncated = false,
 }: AxisRangeControlProps) {
   const validator = useCallback((range?: NumberOrDateRange): {
     validity: boolean;
@@ -60,10 +62,10 @@ export default function AxisRangeControl({
         allowPartialRange={false}
         containerStyles={containerStyles}
         validator={validator}
-        //DKDK
+        //DKDK defaultIndependentAxisRange
         defaultIndependentAxisRange={defaultIndependentAxisRange}
-        //DKDK add something that this comes from axis range control
-        isAxisRangeControl={true}
+        //DKDK show popup
+        isAxisTruncated={isAxisTruncated}
       />
     ) : (
       <NumberRangeInput
@@ -73,10 +75,10 @@ export default function AxisRangeControl({
         allowPartialRange={false}
         containerStyles={containerStyles}
         validator={validator}
-        //DKDK
+        //DKDK pass defaultIndependentAxisRange
         defaultIndependentAxisRange={defaultIndependentAxisRange}
-        //DKDK add something that this comes from axis range control
-        isAxisRangeControl={true}
+        //DKDK show popup
+        isAxisTruncated={isAxisTruncated}
       />
     )
   ) : null;
--- a/src/components/plotControls/AxisRangeControl.tsx
+++ b/src/components/plotControls/AxisRangeControl.tsx
@@ -18,6 +18,8 @@ export interface AxisRangeControlProps
   range?: NumberOrDateRange;
   /** function to call when widget updates the range */
   onRangeChange?: (newRange?: NumberOrDateRange) => void;
+  /** DKDK default/initial independent axis range */
+  defaultIndependentAxisRange?: NumberOrDateRange;
 }
 
 export default function AxisRangeControl({
@@ -26,6 +28,8 @@ export default function AxisRangeControl({
   range,
   onRangeChange,
   containerStyles,
+  //DKDK
+  defaultIndependentAxisRange,
 }: AxisRangeControlProps) {
   const validator = useCallback((range?: NumberOrDateRange): {
     validity: boolean;
@@ -56,6 +60,10 @@ export default function AxisRangeControl({
         allowPartialRange={false}
         containerStyles={containerStyles}
         validator={validator}
+        //DKDK
+        defaultIndependentAxisRange={defaultIndependentAxisRange}
+        //DKDK add something that this comes from axis range control
+        isAxisRangeControl={true}
       />
     ) : (
       <NumberRangeInput
@@ -65,6 +73,10 @@ export default function AxisRangeControl({
         allowPartialRange={false}
         containerStyles={containerStyles}
         validator={validator}
+        //DKDK
+        defaultIndependentAxisRange={defaultIndependentAxisRange}
+        //DKDK add something that this comes from axis range control
+        isAxisRangeControl={true}
       />
     )
   ) : null;
--- a/src/components/widgets/NumberAndDateRangeInputs.tsx
+++ b/src/components/widgets/NumberAndDateRangeInputs.tsx
@@ -38,6 +38,10 @@ export type BaseProps<M extends NumberOrDateRange> = {
   showClearButton?: boolean;
   /** Text to adorn the clear button; Default is 'Clear' */
   clearButtonLabel?: string;
+  /** DKDK default/initial independent axis range */
+  defaultIndependentAxisRange?: NumberOrDateRange;
+  /** DKDK check axis range control */
+  isAxisRangeControl?: boolean;
 };
 
 export type NumberRangeInputProps = BaseProps<NumberRange>;
@@ -79,6 +83,10 @@ function BaseInput({
   containerStyles,
   showClearButton = false,
   clearButtonLabel = 'Clear',
+  //DKDK
+  defaultIndependentAxisRange,
+  //DKDK set default isAxisRangeControl is false
+  isAxisRangeControl = false,
 }: BaseInputProps) {
   if (validator && (required || rangeBounds))
     console.log(
@@ -92,11 +100,22 @@ function BaseInput({
   const [isReceiving, setIsReceiving] = useState<boolean>(false);
   const [validationWarning, setValidationWarning] = useState<string>('');
 
+  //DKDK
+  const [truncatedAxisWarning, setTruncatedAxisWarning] = useState<string>('');
+
   // handle incoming value changes
   useEffect(() => {
     setIsReceiving(true);
     setLocalRange(range);
   }, [range]);
+
+  //DKDK
+  console.log('range at NumberAndDateRangeInputs =', range?.min, range?.max);
+
+  // //DKDK does not work
+  // // const span = <span> whatever your string </span>
+  // const truncatedAxisText = "Gray shadow(s) indicates that data is truncated/not shown<br />due to your range selection";
+  // const truncatedAxisTextSpan = `${truncatedAxisText}`
 
   // if we are not currently receiving incoming data
   // pass localRange (if it differs from `range`) out to consumer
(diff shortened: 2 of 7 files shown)


bobular (commented):
Hi DK - great to see this in motion.

I have some general comments to add before you do more work on this, and then maybe later I can be more specific.

Firstly, the current approach has broken the storybook stories - this one `/story/plots-histogram--range-selection` in particular. A fully backwards compatible solution would be prefered :-)

This may be work-in-progress but the size of the grey rectangles is concerning me.  If the raw data goes from 0 to 100 and the user has chosen to view 70-80, this doesn't mean the grey areas should be 0-70 and 80-100.  That would mean there was nothing to gain from "zooming in" - no new screen space is created.

I think a better solution would be to add something like these props to Histogram, Barplot, Scatterplot:

```
  truncationConfig?: {
    independentAxis?: {
       min?: string | boolean;
       max?: string | boolean;
    },
    dependentAxis?: {
       min?: string | boolean;
       max?: string | boolean;
    },
  }
```
So if the client knows the x-axis is truncated on the left hand side, the props should be 
`truncationConfig: { independentAxis: { min: true } }`

(or maybe call it `truncationSpec` or `truncationInfo` or something like that)

If a 'truthy' string is passed, maybe that can be used as mouse-over text.

If an axis is truncated, the plot component should probably add 5% extra range (e.g. to `plotlyIndependentAxisRange` in Histogram.tsx) and then render the grey rectangle there.

Finally, the passing of axis-specific props into the number and date range component(s) makes me uncomfortable.  Let's find a way to avoid this. These components should be as dumb and single-purposed as possible.

moontrip:
@bobular thank you for sharing your thoughts in detail 👍 I guess you saw my comment at the corresponding web-eda PR (https://github.com/VEuPathDB/web-eda/pull/296) so I expected you would leave this in there 😄 Anyway, I leave my in-line comments/questions/answers next to your great comments. (starting with **DK:** for convenience)  

> Hi DK - great to see this in motion.
> 
> I have some general comments to add before you do more work on this, and then maybe later I can be more specific.
> 
> Firstly, the current approach has broken the storybook stories - this one `/story/plots-histogram--range-selection` in particular. A fully backwards compatible solution would be prefered :-)
> 

**DK:** sorry about this as I worked with histogram filter and did not check stories. Will check them out later.

> This may be work-in-progress but the size of the grey rectangles is concerning me. If the raw data goes from 0 to 100 and the user has chosen to view 70-80, this doesn't mean the grey areas should be 0-70 and 80-100. That would mean there was nothing to gain from "zooming in" - no new screen space is created.
> 

**DK:** I misunderstood that the axis should show whole min/max range. Thank you for correcting me!

> I think a better solution would be to add something like these props to Histogram, Barplot, Scatterplot:
> 
> ```
>   truncationConfig?: {
>     independentAxis?: {
>        min?: string | boolean;
>        max?: string | boolean;
>     },
>     dependentAxis?: {
>        min?: string | boolean;
>        max?: string | boolean;
>     },
>   }
> ```
> 
> So if the client knows the x-axis is truncated on the left hand side, the props should be
> `truncationConfig: { independentAxis: { min: true } }`
> 
> (or maybe call it `truncationSpec` or `truncationInfo` or something like that)
> 

**DK:** I might have misunderstood your intention but does client already know the status of truncated axis without this prop as the range control's value is changed? Though, this is not the case for dependentAxis as there is no data's min/max info from backend response: this is probably the reason why there is no initial min/max range in the dependent axis range control.

> If a 'truthy' string is passed, maybe that can be used as mouse-over text.
> 

**DK:** I wonder where the mouse-over event should be occurred? In the grey areas?

> If an axis is truncated, the plot component should probably add 5% extra range (e.g. to `plotlyIndependentAxisRange` in Histogram.tsx) and then render the grey rectangle there.
> 

**DK:** I will check the percentage and see if it needs to be a bit larger for better display purpose. While this is doable for independent axis (X-axis), as I mentioned in [the comment (web-eda PR)](https://github.com/VEuPathDB/web-eda/pull/296#issue-707489018), dependent axis (y-axis) seems not to be done in this manner because it replies on layout.yaxis.range directly, not backend.

> Finally, the passing of axis-specific props into the number and date range component(s) makes me uncomfortable. Let's find a way to avoid this. These components should be as dumb and single-purposed as possible.

**DK:** Well I got an inspiration how you did treat validation of user's range inputs 😄 Anyway, the main reasons why I passed the props are two folds: a) avoiding overlap with green warning message; b) specifying message for X-axis only. There may be a better way to handle this though.

moontrip:
@bobular I have modified the size of truncated axis to be 5 % of data range. Took more time than I expected because of the case of date string etc. Note that I did not clean up unnecessary codes yet as this is still WIP.

bobular:
Hi @moontrip - the 5% grey bars look and work fine, but the implementation is problematic in a couple of ways.  The first issue is that the variables `lowerX0`, `upperX1` and `upperX0` are modified in the memo'ed `truncatedAxisHighlighting` calculation. The likelihood of unintended side-effects seems risky and creates code that is difficult to read and maintain. Also the naming of these variables is not self-explanatory.

To answer your question above, _yes - the client will decide if and when the truncated warning graphics need to be shown_.  The plot components should just be dumb and do what they are told.  The plot component should only render a warning graphic when needed (second issue: the current WIP implementation is rendering zero width grey rectangles when they are not needed - this could cause mouse-over issues later on?). 

The implementation should go something like this

- two conditions should be met if a grey border will be shown (example given for dependentAxis upper limit)
  - a prop such as `truncationSpec.dependentAxis.max` should be truthy 
  - `dependentAxisRange` (min and max) should be defined
- the plot component then adds 5% extra range to the range that gets passed to Plotly's `layout.yaxis.range`
   - we could have a reusable function for this: `plotlyAxisRange(range : NumberOrDateRange, padStart : number, padEnd : number, logscale : number) : Layout['xaxis']['range']` which could be used in XYPlot too (that also has 5% padding IIRC).  Given an undefined input range, it should return `[undefined, undefined]`.
- a warning rectangle is also added to `layout.shapes`

The independent axis would be treated exactly the same way.

moontrip:
@bobular Thank you for your detailed comment! This reminds me again that this ticket is way more complex than what the title suggests 😄 My comments can be found in the following:

> Hi @moontrip - the 5% grey bars look and work fine, but the implementation is problematic in a couple of ways. The first issue is that the variables `lowerX0`, `upperX1` and `upperX0` are modified in the memo'ed `truncatedAxisHighlighting` calculation. The likelihood of unintended side-effects seems risky and creates code that is difficult to read and maintain. Also the naming of these variables is not self-explanatory.
> 

**DK:** I don't think lowerX0 etc. would be changed inside `truncatedAxisHighlighting`: if it happens, it means `truncatedAxisHighlighting` is also changed due to 2nd argument (dependencies) of useMemo hook. All lowerX0 etc. are affected by the current argument. However, it may be better to place them out of `truncatedAxisHighlighting` - in this case, other variables such as range and rightCoordinate should also be out so I need to think of it whether it has more advantages. I don't like the naming of the variables either as they are set to be temporary. Perhaps I can use a bit longer name, like truncatedAxisLowerX0 etc.

> To answer your question above, _yes - the client will decide if and when the truncated warning graphics need to be shown_. The plot components should just be dumb and do what they are told. The plot component should only render a warning graphic when needed (second issue: the current WIP implementation is rendering zero width grey rectangles when they are not needed - this could cause mouse-over issues later on?).

**DK:** yes you are right. they are technically zero rectangles. So, you indirectly answered one of my previous question, the location of the mouseover event. As a technical aspect, it seems that we need to use `onUpdate` or `onAfterPlot` event like legend tooltip, but I am not quite sure if it is possible in this case, especially considering the fact that those rectangles, `layout.shapes`, are not unique for histogram. That is, selected range is also made of `layout.shapes` and they would probably be not distinguishable each other: they are all generated by `svg.path`. 

> 
> The implementation should go something like this
> 
> * two conditions should be met if a grey border will be shown (example given for dependentAxis upper limit)
>   
>   * a prop such as `truncationSpec.dependentAxis.max` should be truthy
>   * `dependentAxisRange` (min and max) should be defined
> * the plot component then adds 5% extra range to the range that gets passed to Plotly's `layout.yaxis.range`
>   
>   * we could have a reusable function for this: `plotlyAxisRange(range : NumberOrDateRange, padStart : number, padEnd : number, logscale : number) : Layout['xaxis']['range']` which could be used in XYPlot too (that also has 5% padding IIRC).  Given an undefined input range, it should return `[undefined, undefined]`.
> * a warning rectangle is also added to `layout.shapes`
> 
> The independent axis would be treated exactly the same way.

**DK:** The independent axis is much more straightforward than dependent axis case. I will think of this part more carefully and reply in the other comment later

bobular:
Re: variable naming

I'm surprised and confused that there are only three variables.  Surely there should be four, named something like this?

```
independentAxisLeftExtensionStart
independentAxisLeftExtensionEnd
independentAxisRightExtensionStart
independentAxisRightExtensionEnd
```

bobular:
Also just a reminder that we (I've checked with @dmfalke and @steve-fischer-200) would like `defaultIndependentAxisRange` removed and replaced with the four explicit props that turn on each axis extension warning separately. This should simplify the plot component implementation significantly.  Thanks!

bobular:
> 
> 
> > Re: variable naming
> > I'm surprised and confused that there are only three variables. Surely there should be four, named something like this?
> > ```
> > independentAxisLeftExtensionStart
> > independentAxisLeftExtensionEnd
> > independentAxisRightExtensionStart
> > independentAxisRightExtensionEnd
> > ```
> 
> @bobular Thank you for your suggestion on the naming! The reason why independentAxisLeftExtensionEnd was not used is that it simply corresponds to range.min but I will use your suggestion.

OK, cool.  You can keep 'lower' and 'upper' instead of 'left' and 'right' if you prefer (that will make things more consistent for the dependent axis)

moontrip:
> Also just a reminder that we (I've checked with @dmfalke and @steve-fischer-200) would like `defaultIndependentAxisRange` removed and replaced with the four explicit props that turn on each axis extension warning separately. This should simplify the plot component implementation significantly. Thanks!

@bobular It would be great if you can answer me about the following questions: a) I am confused about what the four explicit props mean. Does it like default min/max for each axis, e.g., defaultIndependentAxisMin, defaultIndependentAxisMax, defaultDependentAxisMin, and defaultDependentAxisMax ? b) I thought we don't need to show truncated dependent axis for min (lower bound), which would be a bit strange to see it at a plot. But I just want to make sure if we will have it or not.

bobular:
> > Also just a reminder that we (I've checked with @dmfalke and @steve-fischer-200) would like `defaultIndependentAxisRange` removed and replaced with the four explicit props that turn on each axis extension warning separately. This should simplify the plot component implementation significantly. Thanks!
> 
> @bobular It would be great if you can answer me about the following questions: a) I am confused about what the four explicit props mean. Does it like default min/max for each axis, e.g., defaultIndependentAxisMin, defaultIndependentAxisMax, defaultDependentAxisMin, and defaultDependentAxisMax ? b) I thought we don't need to show truncated dependent axis for min (lower bound), which would be a bit strange to see it at a plot. But I just want to make sure if we will have it or not.

Hi @moontrip 

The four props (though I suggest nesting them in a single prop) are flags. Each prop tells the component if an axis has been truncated.  That is all.  (Although I had an idea that truthy string values might also contain a message to display - we don't need this right now. May never need it.)

E.g. if `truncationSpec.dependentAxis.min = true` then the plot component should indicate that that the lower part of the y-axis is truncated.  I agree with b) that we may not want a large rectangle in that [particular spot](https://docs.google.com/presentation/d/17lZaJShZVUPiBthTwnIKd0Jf-3GZRJ4aTNavIrl0onQ/edit#slide=id.ge3b4530c6a_2_208) (but I don't see any harm in trying), but the component does need to display something.

bobular:
Please also note that the plot component can only draw the warning graphics if it is given the explicit axis range (e.g. `dependentAxisRange` or `independentAxisRange` - with both min and max defined). If the flag is true but the appropriate range isn't provided, I suggest to log a console error.  But in practice that should never happen - when a user zooms, a fully defined xxAxisRange is passed to the plot component.


As describe here:
https://github.com/VEuPathDB/web-components/pull/186#issuecomment-897557690

moontrip:
@bobular thank you for your detailed explanations and suggestions 👍 It would be of great help to move on.

moontrip:
@bobular I have significantly changed this work following your suggestions and fixing several issues I found: Zulu time zone is a part of this work 😄 

Although there are lots of things to talk in this commit, I just describe something off the top of my head - due to this reason, I may miss something haha:

a) changed the rectangular color to be lighter to distinguish from All (not subsetted) case (see the 3rd screenshot)
b) tried to use reusable functions and props as many as possible, but their reusability need to be check when working with other applications - some may need further refinements
c) by introducing `truncationConfig`, a broken histogram story you mentioned (range selection) is now fixed - histogram viz as well 
d) truncation for lower bound of dependent axis is also included, just in case.
e) log scale is also considered, however, it looks a bit strange due to the nature of a log scale plot - i.e., lower bound of dependent axis is overrated, while upper bound still exists but not that visible.

I have also checked mouseover event from your finding (using add_polygons), but unfortunately, it is for R-language only, not plotly.js. I examined other possibility, but could not find appropriate way. One trick I found is to introduce scatter plot on top of those truncated regions, but I am not quite sure if it is plausible given our current implementation, as it would probably cause other side effects.

Some screenshots (number, date, and filtered number cases in order) are shown as follows:

truncated-number1

truncated-date1

truncated-number-filter1

moontrip:
new works are made at the new PR, #195, thus closing this PR

Issue #127, opened 2021-05-21
General: Qualitative colormap


asizemore:
Based on Tol's muted colormap, we propose to use the following 8 colors in order as the qualitative colormap:

rgb(136,34,85)
rgb(136,204,238)
rgb(153,153,51)
rgb(51,34,136)
rgb(68,170,153)
rgb(221,204,119)
rgb(204,102,119)
rgb(17,119,51)

asizemore:
Open question: What should we do if there is no overlay var? 

Options (from Danielle):
1. Use the first color in the tol palette
2. Use the same color as the subset tab
3. Use a color from the site palette
4. Use black
5. Something else?

dmfalke:
> Open question: What should we do if there is no overlay var?

I think the decision was to use the first color in the palette.

dmfalke:
I think we will need to allow for colors to be specified when configuring a plot. This will support use cases where we want to use the same color for values across different plots. My suggestion is that we abstract this by referring to colors by their index number in the palette array. This will allow us to simply have a reference array of values, and we can use the index of that array to specify a palette color.

bobular:
Prelim task: find out what plotly does with empty traces.  So if three traces are normally coloured red,green,blue, and you provide, trace1,emptytrace,trace3, do they still get coloured red and blue (not red,green)

d-callan:
rgb(221,221,221) is the compatible gray, needed for 'No data' values in strata vars. in scatter, i think we also want to use an X marker for 'No data' values to ensure ada compliance if we start using gradient scales there.

dmfalke:
I wonder if we want to have an explicit noData property for our various
plots. I think this will make it easier to ensure consistency, both in
colors and markers. If we go this way, the gray should not be included as a
part of the colorway we pass to plotly.

aaronwlsong:
**QA checking**
Referred  ticket #187, checked all applied plots. On QA website, without overlay variable, the color is the first one: rgb(136,34,85). 

Scatter
<img width="1552" alt="WeChat8062ca86348cfb16558749addac39da8" src="https://user-images.githubusercontent.com/61077845/130123895-4b14f40f-ed4e-41eb-af9f-17abf2af7e8d.png">
 
Histogram
![image](https://user-images.githubusercontent.com/61077845/130124083-4756a05f-4d0a-40bb-8e2b-3c9c7a719cab.png)

Barplot
[...]

PR #187, opened 2021-08-12
Add color palette default and prop

chowington:
Fixes https://github.com/VEuPathDB/web-components/issues/127

The color palette works for the following plots:

- Barplot
- Boxplot
- Histogram
- XYPlot

and doesn't work for the following plots:

- Heatmap (not qualitative)
- Mosaic
- Pie

### DIFF
--- a/src/plots/PlotlyPlot.tsx
+++ b/src/plots/PlotlyPlot.tsx
@@ -133,6 +133,7 @@ export default function PlotlyPlot<T>(
       displayLegend,
       legendTitle,
       title,
+      colorPalette,
     ]
   );
 
--- a/src/plots/PlotlyPlot.tsx
+++ b/src/plots/PlotlyPlot.tsx
@@ -8,7 +8,12 @@ import React, {
 import { PlotParams } from 'react-plotly.js';
 import { legendSpecification } from '../utils/plotly';
 import Spinner from '../components/Spinner';
-import { PlotLegendAddon, PlotSpacingAddon } from '../types/plots/addOns';
+import {
+  PlotLegendAddon,
+  PlotSpacingAddon,
+  ColorPaletteAddon,
+  ColorPaletteDefault,
+} from '../types/plots/addOns';
 import { LayoutLegendTitle } from '../types/plotly-omissions';
 // add d3.select
 import { select } from 'd3';
@@ -52,7 +57,7 @@ const Plot = lazy(() => import('react-plotly.js'));
  *
  */
 export default function PlotlyPlot<T>(
-  props: Omit<PlotProps<T>, 'data'> & PlotParams
+  props: Omit<PlotProps<T>, 'data'> & PlotParams & ColorPaletteAddon
 ) {
   const {
     title,
@@ -68,6 +73,7 @@ export default function PlotlyPlot<T>(
     maxLegendTextLength = 20,
     // expose data for applying legend ellipsis
     data,
+    colorPalette,
     ...plotlyProps
   } = props;
 
@@ -118,6 +124,7 @@ export default function PlotlyPlot<T>(
         ...(legendOptions ? legendSpecification(legendOptions) : {}),
       },
       autosize: true, // responds properly to enclosing div resizing (not to be confused with config.responsive)
+      colorway: colorPalette ?? ColorPaletteDefault,
     }),
     [
       plotlyProps.layout,
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -9,6 +9,7 @@ import {
   OrientationDefault,
   DependentAxisLogScaleAddon,
   DependentAxisLogScaleDefault,
+  ColorPaletteAddon,
 } from '../types/plots';
 import PlotlyPlot, { PlotProps } from './PlotlyPlot';
 import { Layout } from 'plotly.js';
@@ -20,7 +21,8 @@ export interface BarplotProps
     BarLayoutAddon<'overlay' | 'stack' | 'group'>,
     OrientationAddon,
     OpacityAddon,
-    DependentAxisLogScaleAddon {
+    DependentAxisLogScaleAddon,
+    ColorPaletteAddon {
   /** Label for independent axis. e.g. 'Country' */
   independentAxisLabel?: string;
   /** Label for dependent axis. Defaults to 'Count' */
--- a/src/plots/Boxplot.tsx
+++ b/src/plots/Boxplot.tsx
@@ -8,6 +8,7 @@ import {
   OpacityDefault,
   OrientationAddon,
   OrientationDefault,
+  ColorPaletteAddon,
 } from '../types/plots';
 import { NumberOrDateRange } from '../types/general';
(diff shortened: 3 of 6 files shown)


src/plots/PlotlyPlot.tsx
```diff
@@ -118,6 +124,7 @@ export default function PlotlyPlot<T>(
         ...(legendOptions ? legendSpecification(legendOptions) : {}),
       },
       autosize: true, // responds properly to enclosing div resizing (not to be confused with config.responsive)
+      colorway: colorPalette ?? ColorPaletteDefault,
     }),
     [
       plotlyProps.layout,
```
jtlong3rd, line 130: Add `colorPalette` to this `useMemo` dependency array.

jtlong3rd (commented):
Aside from one small suggestion, looks good to me!

dmfalke (changes requested):
This is looking good. I think XYPlot does work, but the colors are hard-coded in `XYPlot.stories.tsx`. Could you add a story without the hard-coded colors? Thanks!

src/plots/PlotlyPlot.tsx
chowington: Done!

chowington:
> 
> 
> This is looking good. I think XYPlot does work, but the colors are hard-coded in `XYPlot.stories.tsx`. Could you add a story without the hard-coded colors? Thanks!

Done! Looks like the color palette gets used by the dots and the smoothed mean, but not the confidence interval

moontrip:
@chowington for scatter plot, there exists a sort of a grouped plot like raw data (marker), smoothed mean, and confidence interval (CI) for smoothed mean with raw option, which would follow the same color: CI uses opacity lower than 1 but the same color with corresponding marker and line (smoothed mean).

jtlong3rd (approved):

src/plots/PlotlyPlot.tsx
```diff
@@ -68,6 +73,7 @@ export default function PlotlyPlot<T>(
     maxLegendTextLength = 20,
     // expose data for applying legend ellipsis
     data,
+    colorPalette,
```
bobular, line 76: I'd have put the default here - 

```
colorPalette = ColorPaletteDefault,
```
Just in case it gets used more than once and we forget to `colorPalette ?? ColorPaletteDefault` each time we use it.

But this was only ever going to be used once - so no need for a post-merge change.  Maybe next time we're working on this file we could change it.

chowington: Right, of course I should have! That's the second duh moment in two days for me, lol. Should have added you as the third reviewer! Thanks Bob

bobular:
It is working for Mosaic, isn't it? @chowington 

image

Pie uses the colorway too if you disable `defaultColorGen` (I made it `yield undefined`)

image


I'm confused about one thing: Barplot, Histogram and Boxplot all use the ColorPaletteAddon but XYPlot, Pie and Mosaic don't.
Ah but then I see it's here in PlotlyPlot.tsx, which makes sense
```
export default function PlotlyPlot<T>(
  props: Omit<PlotProps<T>, 'data'> & PlotParams & ColorPaletteAddon
) {
```
Does that mean we should remove the addon from Barplot, Histogram and Boxplot?

bobular:
> @chowington for scatter plot, there exists a sort of a grouped plot like raw data (marker), smoothed mean, and confidence interval (CI) for smoothed mean with raw option, which would follow the same color: CI uses opacity lower than 1 but the same color with corresponding marker and line (smoothed mean).

You get the desired cyan colour for the smoothed mean CI with the following minor tweak to the story

```
diff --git a/src/stories/plots/XYPlot.stories.tsx b/src/stories/plots/XYPlot.stories.tsx
index 303f257..ac89213 100755
--- a/src/stories/plots/XYPlot.stories.tsx
+++ b/src/stories/plots/XYPlot.stories.tsx
@@ -1053,7 +1053,7 @@ function processInputData<T extends number | string>(
         // type: 'line',
         type: 'scattergl',
         // here, line means upper and lower bounds
-        line: { color: 'transparent', shape: 'spline' },
+        line: { color: defineColors ? 'transparent' : undefined, shape: 'spline' },
       });
     }
```

moontrip:
> > @chowington for scatter plot, there exists a sort of a grouped plot like raw data (marker), smoothed mean, and confidence interval (CI) for smoothed mean with raw option, which would follow the same color: CI uses opacity lower than 1 but the same color with corresponding marker and line (smoothed mean).
> 
> You get the desired teal/cyan colour for the smoothed mean CI with the following minor tweak to the story
> 
> ```
> diff --git a/src/stories/plots/XYPlot.stories.tsx b/src/stories/plots/XYPlot.stories.tsx
> index 303f257..ac89213 100755
> --- a/src/stories/plots/XYPlot.stories.tsx
> +++ b/src/stories/plots/XYPlot.stories.tsx
> @@ -1053,7 +1053,7 @@ function processInputData<T extends number | string>(
>          // type: 'line',
>          type: 'scattergl',
>          // here, line means upper and lower bounds
> -        line: { color: 'transparent', shape: 'spline' },
> +        line: { color: defineColors ? 'transparent' : undefined, shape: 'spline' },
>        });
>      }
> ```

@bobular just FYI: certainly it would show teal/cyan color but the main purpose of the `line:` prop in there is not to show upper/lower bound lines: if it is defined like that, it would shows bound lines and a vertical line at the max of confidence interval

bobular:
Ah yes it does show boundary lines in the story, but not in the
ScatteplotViz so I don't think we need to worry.

chowington:
> 
> 
> It is working for Mosaic, isn't it? @chowington
> 
> image
> 
> Pie uses the colorway too if you disable `defaultColorGen` (I made it `yield undefined`)
> 
> image
> 
> I'm confused about one thing: Barplot, Histogram and Boxplot all use the ColorPaletteAddon but XYPlot, Pie and Mosaic don't.
> Ah but then I see it's here in PlotlyPlot.tsx, which makes sense
> 
> ```
> export default function PlotlyPlot<T>(
>   props: Omit<PlotProps<T>, 'data'> & PlotParams & ColorPaletteAddon
> ) {
> ```
> 
> Does that mean we should remove the addon from Barplot, Histogram and Boxplot?
> 
> EDIT:
> I moved the ColorPaletteAddon usage in PlotlyPlot.tsx and now none of the individual components need to use it.
> 
> ```
> export interface PlotProps<T> extends ColorPaletteAddon {
> ```

Hi Bob! I made ColorPalette an Addon because not all plots used it (at this point only Heatmap doesn't), therefore it should not be provided as a prop for all plots. So I left it out of the PlotProps interface. Adding it to that interface means that Heatmap has it as a prop, which is confusing. However the implementation was the same for all plots that used ColorPalette, so I just put the implementation in PlotlyPlot by extending the actual PlotlyPlot incoming props type directly.

bobular:
Aha, thanks for the explanation, @chowington - so yes, we should move it out of PlotlyPlot's props into each of the components.  A minor faff, but it makes sense for heatmap and any future plots that don't use it.  I'll make a ticket directly in web-components.

PR #190, opened 2021-08-18
Simplification of Histogram x-axis zooming and range selection fixes

bobular:
Working on a simple Story for "truncated axis" stuff, I found that Histogram had a stupid feature where it wouldn't allow you to zoom in on the x-axis and thus hide some of the data that was given to the component.  I think partly the intention was to avoid chopping off bins (in the x-axis direction).  Histogram (and other plots) will have new truncation warnings to make this clear.

There was also a complicated modification to the first and last bins' `binStart` and `binEnd` for range selection purposes, involving a constraint prop called `selectedRangeBounds`.  People have commented that we don't need to constrain the selected range bounds, and I agree. 

So in this ticket I have

1. modified `SelectedRangeControl` to continue to take the `selectedRangeBounds` prop (for auto-fill purposes, when the user types just one number in), but no longer enforces the range bounds (you can type in age: -100 to +1000 if you want)
2. deprecated `selectedRangeBounds` (and related `isZoomed`) in `Histogram` - it was used to modify selected ranges that included the first or last bins.  It really wasn't worth all the complication.
3. introduced a "rounding" up or down of x-axis range coordinates so that no bin is ever chopped off when user zooms in
4. added a Histogram story with some simple static data - it demonstrates the "rounding" on the x-axis range: type 2.1 to 2.9 and it will make sure that you get the whole 2.0 to 3.0 bin.
5. selections on date variables: the end date has 1 day subtracted - see https://github.com/VEuPathDB/web-eda/pull/315

### DIFF
--- a/src/components/plotControls/SelectedRangeControl.tsx
+++ b/src/components/plotControls/SelectedRangeControl.tsx
@@ -39,7 +39,15 @@ export default function SelectedRangeControl({
     : // use a custom validator when we don't want to
       // actually constrain the range between the bounds.
       useCallback((range?: NumberOrDateRange) => {
-        if (range && range?.min > range?.max) {
+        if (
+          range &&
+          (valueType === 'date'
+            ? // only compare the date part (one can be datetime, just-entered value is just a date)
+              (range.min as string).substring(0, 10) >
+              (range.max as string).substring(0, 10)
+            : range?.min > range?.max)
+        ) {
+          console.log(range);
           return {
             validity: false,
             message: 'Range start cannot be above range end',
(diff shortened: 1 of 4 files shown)


moontrip:
@bobular honestly I don't know well about bin-related stuffs nor bugs you mentioned so I focused on whether it worked fine. I have tested both web-component and web-eda and it appears to me that they worked well 👍

bobular:
> @bobular honestly I don't know well about bin-related stuffs nor bugs you mentioned so I focused on whether it worked fine. I have tested both web-component and web-eda and it appears to me that they worked well +1

You're right - it's all quite dense code and there are details that depend on back end behaviours (are filters inclusive/exclusive of the start/end dates etc).  I have hopefully removed some of the complexity and thank you for testing!

dmfalke (approved):

moontrip (approved):

Issue #182, opened 2021-08-06
Visualizations: swap count and percent data in hover over for RxC and 2x2


SheenaTomko:
Swap the location of count and % in the hover over. Usually on tables in epidemiology papers, data is presented as N (%)

SheenaTomko:
looks good

Issue #181, opened 2021-08-06
Visualizations: Move x-axis labels on RxC and 2x2 plots


SheenaTomko:
Move X-axis labels below the axis and don’t show the % on bottom. Also, remove the % and count from the end of the labels.

chowington:
What's the reason for removing the raw counts and percentages from the x-axis values at the bottom? To me it seems like useful information that can't be gotten anywhere else from the plot or table view (no percentages). I know that's not how the original Shiny plot is or probably how contingency tables are generally. Maybe we could make it an option to turn them on via the plot controls at the bottom of the page (visible on other visualizations)? I have the same feeling about the 0-100% x-axis scale. Thoughts?

SheenaTomko:
Part of the reason to remove the scale from the x-axis is just so it doesn't get in the way of the value labels. I would be ok with an option to turn it on via plot controls at the bottom.

For the x-axis value labels themselves, I think it's a little confusing to have those counts and percentages for that variable when we don't do the same for y-axis values. It also makes the labels longer and harder to read. In theory, we could add that information to the y-axis labels as well, but I think it might be overkill. 

@d-callan can add her thoughts

d-callan:
to me, keeping the treatment of the axes as consistent with each other as possible, and with the general expectations for a mosaic plot is important. mosaic plots usually have categorical axes labels, and not a scale representing the percentage. a user familiar w mosaic plots will potentially be confused by this, and one who isnt shouldnt be bothered either way. i tend to think we shouldnt offer it, even as an option. as for the % in the labels, i agree w sheena they feel a bit like theyre cluttering the space, but would be fine w offering those as an option.

chowington:
Thanks for your comments both! Since there doesn't seem to be consensus or a big drive to keep these extra elements, I'll just go ahead and remove them both as requested at first for now.

SheenaTomko:
Looks good thanks. Going to open a second ticket to add the variable name as an axis label

PR #191, opened 2021-08-19
Various mosaic label changes

chowington:
- Swap percent and raw value positions in hover labels
- Remove percents and raw values from x-axis labels
- Fix bug where x-axis labels were showing even when
showColumnLabels was false

Edit for related tickets:
- https://github.com/VEuPathDB/web-components/issues/181
- https://github.com/VEuPathDB/web-components/issues/182

### DIFF
--- a/src/plots/MosaicPlot.tsx
+++ b/src/plots/MosaicPlot.tsx
@@ -46,13 +46,8 @@ export default function MosaicPlot({
       tickvals: column_centers,
       ticktext:
         showColumnLabels !== false
-          ? data.independentLabels.map(
-              (value, i) =>
-                `<b>${value}</b> ${percent_widths[i].toFixed(1)}% (${raw_widths[
-                  i
-                ].toLocaleString('en-US')})`
-            )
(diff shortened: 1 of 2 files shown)


jtlong3rd (approved):

bobular (commented):
I like the styling 
image

But after I click on a tab heading in Firefox, it stays greyed out and I'm not sure that was intentional.  Going to test in Chrome now...

image

Chrome results: same thing.  

If I click and mouse-out of the tab header ASAP - the grey shading goes away (Chrome + FF) - but we cannot expect users to be so ninja.

bobular (approved):
I'll approve here because it seems tab-related stuff is on the EDA side?

chowington:
@bobular This PR is not related to tabs--it deals with these tickets:

- https://github.com/VEuPathDB/web-components/issues/181
- https://github.com/VEuPathDB/web-components/issues/182

Thanks for the feedback though! There's not a ticket yet for styling the tabs, so I don't know quite where your comment should go

PR #193, opened 2021-08-24
Birds eye plot component

bobular:
Here we have, I think, a pretty final Birds Eye View plot component.

Perhaps the only shortfall is that the empty component has less height than the full component.  We would need to add at least one dummy `CountBox` element to claim that space? (EDIT: I tried that and pushed the changes)

Also, is the mouse-over `visibility` implementation efficient enough?

I would be grateful for suggestions.

Also: styling is a bit _ad hoc_ as usual!

Also also: I gave Dave a rest from PR reviewing ;-) 

image


jtlong3rd (approved):

bobular:
@danicahelb @d-callan @asizemore 
I went for line width 2 for the "brackets".
With line width 1 I thought they were not distinguishable enough from the newly added y-axis baseline (which helped reduce Dave's confusion).
Are you OK with this?
image

d-callan:
I'm happy enough to see it in context before passing any further judgements.

asizemore:
Same! It looks really good! I also like the "Has data for" change -- very clear.

PR #194, opened 2021-08-24
Make mosaic box outlines smaller

chowington:


chowington:
Merging because it's a trivial change

PR #201, opened 2021-08-27
PlotRef (React.Ref type)

dmfalke:
This PR adds the ability for a consumer of a plot to generate an data uri for an image. It does so by introducing a generic imperative API (`PlotRef`) for interacting with plots. This API is exposed as a React `ref`, using the `React.forwardRef` utility.

A factory function (`makePlotlyPlotComponent`) has been added for creating a React component that uses `PlotlyPlot`. It encapsulates the machinery needed to expose the `PlotRef` interface mentioned above.

This will be needed by https://github.com/VEuPathDB/web-eda/issues/272


dmfalke:
I've added a utility (`SharedPromise`) to address the issue mentioned above.

jtlong3rd (approved):

PR #202, opened 2021-08-30
legend title ellipsis with tooltip

moontrip:
This PR addressed [the issue at the web-eda, #332](https://github.com/VEuPathDB/web-eda/issues/332)

Tried to wrap lengthy legend title text at first. However it was found that the area of legend title did not allow multiple lines: wrapped texts were overlapped in the single line.

Thus, it was decided to use ellipsis with tooltip. A screenshot is attached as follows:

legend-title-ellipsis

### DIFF
--- a/src/plots/PlotlyPlot.tsx
+++ b/src/plots/PlotlyPlot.tsx
@@ -192,7 +192,7 @@ export default function PlotlyPlot<T>(
           .text(legendTitle);
       }
     },
-    [storedLegendList]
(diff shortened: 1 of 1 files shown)


jtlong3rd (approved):

moontrip:
Thank you for your quick review @jtlong3rd 👍

moontrip:
@bobular thank you for your suggestion. Added as follows: now you can point out the location, if necessary 😄

PR #203, opened 2021-09-01
Make entity diagram use plural entity names

chowington:


bobular (approved):
Looks good.  I tried a few studies and didn't notice any that overflowed their boxes.

dmfalke (approved):

PR #204, opened 2021-09-02
define ColorPaletteDark at addOns

moontrip:
@dmfalke here is the change to have ColorPaletteDark at addOns.ts as we discussed.

### DIFF
--- a/src/types/plots/addOns.ts
+++ b/src/types/plots/addOns.ts
@@ -114,6 +114,18 @@ export const ColorPaletteDefault: string[] = [
   'rgb(17,119,51)',
 ];
 
+/** Darker color pallette for Scatter plot viz for now */
+export const ColorPaletteDark: string[] = [
+  'rgb(115, 28, 72)',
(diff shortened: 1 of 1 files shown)


src/types/plots/addOns.ts
```diff
@@ -114,6 +114,18 @@ export const ColorPaletteDefault: string[] = [
   'rgb(17,119,51)',
 ];
 
+/** Darker color pallette for Scatter plot viz for now */
```
dmfalke, line 117:
```suggestion
/** Darker color palette, useful for overlay traces, such as smoothed mean in scatter plots  */
```

dmfalke (changes requested):
Looks good, but please update the comment. Thanks!

src/types/plots/addOns.ts
moontrip: @dmfalke Thank you for your suggestion. I have changed it

dmfalke (approved):

PR #206, opened 2021-09-07
follow-ups for truncated axis - warning icon, plot line

moontrip:
This is for follow-up items regarding truncated axis stuffs: see the [previous PR](https://github.com/VEuPathDB/web-eda/issues/98) and https://github.com/VEuPathDB/web-eda/issues/382. Will also make corresponding PR at web-eda.

I have discussed with Sheena about the appropriateness of new warning icon and she said it looked great.

histogramfilter-warning1

Also here are additional screenshot for X-axis line (depending on the presence of y-axis range, esp. yMin > 0 or not)

truncated-xaxis1

truncated-xaxis2

### DIFF
--- a/src/plots/Histogram.tsx
+++ b/src/plots/Histogram.tsx
@@ -352,10 +352,7 @@ const Histogram = makePlotlyPlotComponent(
       automargin: true,
       showgrid: false,
       zeroline: false,
-      showline:
-        orientation === 'vertical'
-          ? !axisTruncationConfig?.dependentAxis?.min
-          : !axisTruncationConfig?.independentAxis?.min,
+      showline: !axisTruncationConfig?.dependentAxis?.min,
       title: {
         text: independentAxisLabel,
       },
@@ -428,7 +425,7 @@ const Histogram = makePlotlyPlotComponent(
             extendedDependentAxisRange?.max,
           ].map((val) =>
             dependentAxisLogScale && val != null
-              ? val < 0
+              ? val <= 0
                 ? 0
                 : Math.log10(val as number)
               : val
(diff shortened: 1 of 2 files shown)


jtlong3rd (approved):

src/plots/Histogram.tsx
```diff
@@ -352,7 +352,10 @@ const Histogram = makePlotlyPlotComponent(
       automargin: true,
       showgrid: false,
       zeroline: false,
-      showline: false,
+      showline:
+        orientation === 'vertical'
```
bobular, line 356: I'm surprised/confused that the orientation switch is needed.

At line 350 we define this as `independentAxisLayout` - it's for the independent axis, regardless of h/v orientation.

So surely the line should only be drawn if `!axisTruncationConfig?.dependentAxis?.min`

bobular (commented):
Looks good - but please comment on the orientation thing!

bobular:
Sorry DK - nearly forgot to post this screenshot - could you take another look at log-scale compatibility please?

image

src/plots/Histogram.tsx
moontrip: @bobular yes you are right. I will slim them down

moontrip:
> Sorry DK - nearly forgot to post this screenshot - could you take another look at log-scale compatibility please?
> 
> image

@bobular thank you for finding the bug. I couldn't find such a bug from my tests before so I will take a look at it 👍

src/plots/Histogram.tsx
moontrip: @bobular addressed this following your suggestion in the new commit

moontrip:
> Sorry DK - nearly forgot to post this screenshot - could you take another look at log-scale compatibility please?
> 
> image

@bobular I finally found the reason why it happened. For the dependent axis, I did set a condition for log scale to check whether the value is negative, however, I forgot to consider the case of value = 0 which caused -Infinity in the dependent axis min. Here is the screenshot with the same variable and range control values you have used above

histogramfilter-update1

bobular (approved):

PR #209, opened 2021-09-13
Remove width prop from ED Text

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/388

I opted for removing the defined width from the ED Text component, as it seems with the box layout we're using we'll never want the text to wrap anyway. This should fix the styling for the length of entity names we currently have, but if we get much longer ones in the future, the text will start overflowing the box bounds.


chowington:
Household repeated measures selected:
image

Participant repeated measures (to be renamed soon) selected:
image

dmfalke (approved):

PR #215, opened 2021-09-23
Change Switch style

chowington:
Works toward https://github.com/VEuPathDB/web-eda/issues/330

In addition to changing switch coloring, also decapitalizes button text in general.


bobular (approved):
FYI I added a commit to remove the label focus of the range inputs.

PR #216, opened 2021-09-29
long y-axis title handled by ellipsis and tooltip

moontrip:
This PR will address the issue, [#395 at web-eda](https://github.com/VEuPathDB/web-eda/issues/395)

A long y-axis title is now handled by ellipsis and tooltip. Currently the max number of characters for y-axis title is set to 50 based on eye test 😃 

A screenshot is shown as follows: somehow backend data service seems not to work correctly so no plot is made, however, it is just okay to see the implementation anyway. Just FYI, this was captured using d-falke data service, GEMS1A Case Control study, and Mosaic 2x2 table plot.

test-long-yAxis-title

### DIFF
--- a/src/plots/PlotlyPlot.tsx
+++ b/src/plots/PlotlyPlot.tsx
@@ -199,7 +199,10 @@ function PlotlyPlot<T>(
   }, [data]);
 
   // keep dependent axis title for tooltip text
-  const originalDependentAxisTitle = plotlyProps?.layout?.yaxis?.title;
+  const originalDependentAxisTitle = useMemo(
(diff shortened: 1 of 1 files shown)


jtlong3rd (approved):

bobular (approved):
Looks good.  I wonder if `maxDependentAxisTitleTextLength` ought to be dependent on the plot's height (in `containerStyles.height` if present) but 50 characters seems plenty even for a much bigger plot.  These long variable names are very much edge cases.

It's complicated converting pixels into number of characters and I don't think the code complexity is worth the minimal functional gains!

PR #221, opened 2021-10-04
Fixed two mouse-over issues for histogram and bar plot

### LINKED ISSUE #219: Subsetting: histogram bin edges and mouseover edges are not aligned
In the subsetting histogram, I get the correct mouseover information for the bin if my cursor is placed from the leftmost portion of each bin to the middle of the bin. If the cursor is moved past the middle of the bin, the mouseover information changes to the info for the bin immediately to the right of the bin i am interested in.

Seems like the bin edges and the mouseover edges are offset from each other

### LINKED ISSUE #220: Overlay value part of viz plot hovertext is invisible with new "no data" styling
Slack discussion [here](https://epvb.slack.com/archives/C012ZK4Q5CZ/p1631624380037500)

Affects just barplot and histogram.

Likely solution is to redo all mouse-over hovertext using hovertemplate and make `hovermode: 'closest'` for all plot types.

Suggested cost: low.

How to replicate:

1. Check out `missing-data-styling` branches in eda and components (unless already merged)
2. GEMS1, barplot with variables as in screenshot below (overlay is `Non-piped water source without tap`), switch on "Include households with no data ..."

Screenshot from 2021-09-14 15-35-03

bobular:
Resolves #220 
Resolves #219 

Switched over to `hovermode: 'closest'` and added custom `hoverformat` text and killed two birds with one stone :-)

### DIFF
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -113,6 +113,10 @@ const Barplot = makePlotlyPlotComponent(
                   color: el.borderColor,
                 },
               },
+              hovertemplate: el.value.map(
+                (value: number, index: number) =>
+                  `${value}<br />${el.label[index]}<br />${el.name}<extra></extra>`
+              ),
             };
           } else {
             return {};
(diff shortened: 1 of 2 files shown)


moontrip (approved):
Looks good 👍

Issue #185, opened 2021-08-10
Visualizations: Bar plot log y axis shows repeated tick marks


asizemore:
When in Gems -> no subset -> barplot with Study Year as main variable, after switching to a log y axis we see duplicated ticks for 7,000 and 8,000 (see below). 

A similar situation occurs when Country, Case or Control Participant, and Age group, is selected as the Main variable, but _not_ Study Timepoint.

<img width="856" alt="Screen Shot 2021-08-10 at 7 51 02 AM" src="https://user-images.githubusercontent.com/11710234/128862929-18f23e72-e61f-4fc3-a6c5-ea804997330a.png">

dmfalke:
It looks like this is related to https://github.com/VEuPathDB/web-components/pull/175

asizemore:
Just followed up to see if #175 fixed the problem. Unfortunately seems to still be an issue. Below is Gems, no filter, Case or Control participant. 
Also seeing this behavior with modes = count, proportion, and main var = Country, Age Group, and Household Study Timepoint, but not Study Timepoint.
<img width="899" alt="Screen Shot 2021-09-08 at 9 57 36 AM" src="https://user-images.githubusercontent.com/11710234/132524752-89c8f262-db1e-4379-a3e1-c97022a8c8c0.png">

moontrip:
Hi @asizemore~

I had a chance to test both cases with `qa` and `qa.restricted` and they both seem to work fine, so I will assign this issue to you for your tests. Please see my screenshots.

![study year](https://user-images.githubusercontent.com/12802305/136067466-c8414dea-ec38-4191-881b-37f9a9447edc.png)

![case or control participant](https://user-images.githubusercontent.com/12802305/136067484-0733aa8c-ca78-4456-83c0-828018ed2511.png)

asizemore:
Thanks @moontrip! I was able to reproduce what you have above in qa, but when I toggle to show y axis on a log scale i am still seeing the repeated marks, please see below:
<img width="893" alt="Screen Shot 2021-10-05 at 1 04 18 PM" src="https://user-images.githubusercontent.com/11710234/136069663-9436e5e7-a02d-4eea-9194-64187bb10e67.png">

I think it's only when i toggle the log scale "on" that it becomes an issue. If log scale is off, both Count and Proportion work great. But when log scale in on, i'm seeing repeated y tick labels for both Count and Proportion

moontrip:
[...]

PR #225, opened 2021-10-11
handling yaxis ticks for logscale at barplot

moontrip:
This is related to the issue #185. Corresponding web-eda changes will be made soon.


moontrip:
For now I will merge this. Will think/handle histogram viz case that I have said above at other PR

PR #230, opened 2021-10-13
Birds eye popup placement explorations

bobular:
This is just storybook work.

Added an example that shows the popup being clipped.

Played around with width and plotly margins.

Can be fixed in EDA without this code, but we should merge at some point!


bobular:
Merging without review as this is just storybook stuff

PR #233, opened 2021-10-21
testing plotly version up

moontrip:
Previous plotly version was 1.58.4 with the setting of ^1.54.1 at `package.json`. The latest plotly.js is 2.5.1 so tested with ^2.5.1. The react-plotly wrapper seems to compile the plotly.js defined at `dependencies`. (I made [a technical debt ticket](https://github.com/VEuPathDB/web-eda/issues/627) regarding this)

There were a couple of hiccups concerning the upgrade of plotly version, especially d3-related errors: I used it for legend/axis label ellipsis/tooltip etc. Somehow web-component did not have volta configuration at `package.json`, which also affected a d3 error (d3-array library did not like a lower version of node.js that my VM has used).  So I have added the same configuration used at web-eda. Then I had to install d3 and @types/d3 packages to work things properly. But, by adding d3's type definition, a type error occurred at PlotlyPlot component (Object is unknown error) so I just set it to be `any` type for now. 

I have tested storybooks we have and they seem to work fine without noticeable issue nor an error. Also, tested plots used at web-eda (Subsetting's histogram and Viz plots) using `yarn link` and from my tests, they seem to work well too: understandably I couldn't test all possible use cases but I couldn't identify specific issues from those limited tests.

Thus, I have made a draft PR so that anyone who is interested can test this upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.8.4",
+  "version": "0.8.5",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.57",
+    "@types/d3": "^7.1.0",
     "@types/date-arithmetic": "^4.1.1",
     "@visx/gradient": "^1.0.0",
     "@visx/group": "^1.0.0",
@@ -20,13 +21,14 @@
     "@visx/tooltip": "^1.3.0",
     "@visx/visx": "^1.1.0",
     "bootstrap": "^4.5.2",
+    "d3": "^7.1.1",
     "date-arithmetic": "^4.1.0",
     "debounce-promise": "^3.1.2",
     "latlon-geohash": "^2.0.0",
     "leaflet": "^1.6.0",
     "leaflet-drift-marker": "^1.0.3",
     "luxon": "^1.25.0",
-    "plotly.js": "^1.54.1",
+    "plotly.js": "^2.5.1",
     "re-resizable": "^6.9.0",
     "react": "^16.13.1",
     "react-bootstrap": "^1.3.0",
@@ -99,5 +101,9 @@
     "stats-lite": "^2.2.0",
     "storybook": "^6.3.6",
     "typescript": "4.3.4"
+  },
+  "volta": {
+    "node": "14.15.5",
+    "yarn": "1.22.10"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.8.5",
+  "version": "0.8.8",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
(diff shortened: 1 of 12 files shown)


chowington:
Just messaging to help this PR along.

@bobular Have you been able to/do you want to test it?
@moontrip Have you fixed that histogram labelling issue yet?

DK, I think it makes sense to have this separate PR for the plotly version upgrade, as it's a big step, and it's a different concern than my mosaic PR, although that one does depend on this one.

moontrip:
@chowington I didn't apply [my solution](https://github.com/VEuPathDB/web-components/pull/233#issuecomment-952519125) to this PR yet as I need @bobular's opinion about that - there may be a reason why he did set like that, which I have no idea

bobular:
I've had to merge from `main` to get this to work with latest EDA. But now I'm having problems with the back end (https://localhost:3000/eda-subsetting-service/studies/GEMSCC0003-1) so I'm going to give up and try again in the morning!

I suspect the `showValues` option was intended for use with MapVEu - maybe the enlarged pop-up marker charts?  Or maybe it was added with no real purpose?

Anyway, I'll investigate fully when I can.

bobular:
> @chowington found that plotly.js version at web-eda still has older one v1.58.5 despite yarn link: there was no dependencies of plotly.js at its package.json. Thus, I have tried to add plotly.js as a dependencies at package.json of web-eda and run yarn install.
> 
> `"plotly.js": "^2.5.1",`
> 
> (or `yarn add plotly.js@^2.5.1` at CLI)
> 
> I confirmed that web-eda's plotly.js is updated to have v2.5.1 with this approach.

I'm worried about this!

So the problem is that `yarn link` doesn't force the consumer project (web-eda) to update dependencies?

`yarn install --force` doesn't do it either.

Should we use [yalc](https://github.com/wclr/yalc)?

dmfalke:
I took a quick look at yalc. It seems complicated. I'm not crazy about it making changes to `package.json`.

This might be a tough pill to swallow, but  perhaps we need to add more stories here, to cover cases invoked by web-eda. Ideally, if we upgrade a dependency here, and all stories look good, then all consumers should be unaffected. Worst case scenario is that we publish an update to this library and miss some use case - we would just have to add a story exercising this use case, fix it, and publish an update. In the end, this will add stability to this library, and we don't depend on another repo for testing.

bobular:
I can't get yalc to work as we want anyway!

bobular:
I'll add a story demonstrating `showValues` and add the fix @moontrip suggested (or similar)

bobular:
I went with `text: showValues ? binCounts.map(String) : undefined`  (instead of empty string) just because we've done the same with other plotly parameters.

I added a "Show Values" story and it all seems OK.

bobular (approved):

moontrip:
@bobular thank you for tests 👍

chowington (approved):
Looks good! Thanks @moontrip  and @bobular!

moontrip:
Thank you for your review @chowington. All (including @dmfalke), so is it alright to merge this ? Also, I think I need to make a PR at web-eda to add the new version of plotly.js ?

bobular:
> Thank you for your review @chowington. All (including @dmfalke), so is it alright to merge this ? Also, I think I need to make a PR at web-eda to add the new version of plotly.js ?

Yes and no.  Yes, a small PR will be needed to upgrade @veupathdb/components 

But no I don't think you want to make a direct "yarn add" of plotly at `web-eda` like you did for testing.

Once the new `web-components` package is published at npm, the usual `yarn add @veupathdb/components` will bring in the new version of plotly.

moontrip:
@bobular I will check if it would resolve the case 👍

moontrip:
@bobular I have tried to accommodate new version of web-component after merging this PR, but it still showed older version of plotly.js. As Dave once mentioned before at slack, it might be coming from others. Thus, I think that I just need to add plotly.js with new version at dependencies (package.json at web-eda) to resolve this issue.

PR #235, opened 2021-10-27
toggle text gray for disabled, else black

asizemore:
Resolves web-eda [#574](https://github.com/VEuPathDB/web-eda/issues/574)

When toggles are disabled, their text will now be in MEDIUM_GRAY. Otherwise, their text will be black.

Also considered making the text a dark gray when enabled, to better match the toggle itself. Would require adding a fourth gray to our collection, so unless we have lots of support to add another gray we can just use black.

Screen Shot 2021-11-03 at 2 32 06 PM


asizemore:
@eikonomega do you mind reviewing quickly when you have a moment please? Thanks!!

eikonomega (approved):
Looks good and makes sense! I think @asizemore is right about not using pure black as a color. It is too harsh. Will look into making a broader range of pre-defined grays available in `core-ui`.

Issue #184, opened 2021-08-09
Visualizations: move Y-axis labels in RxC plot


SheenaTomko:
Move Y-axis labels to the left axis rather than in the legend, and center based on the boxes in the first row. May need additional UI work if there are many values for the Y-axis and the first row (X-axis value) doesn't have some of those values, since there will be no box for it. 
Users should not be able to remove one of the values for the axis.

chowington:
Hi Sheena! Do you envision the y-axis labels being to the right of the numbers, to the right of the variable name, or to the left of the variable name?

![image](https://user-images.githubusercontent.com/12377649/130132141-0aaa9da2-52ac-4304-95b7-dee27bc11dc5.png)

There certainly will be an issue with boxes in the first column that are too small for either the label to fit or the color to be visible. I don't think we can lose the square color indicators present in the current legend. These are going to be challenges from an implementation perspective as well.

SheenaTomko:
Yeah there are definitely challenges here and @d-callan can weigh in as well. I think we want something like this where we loose the y-axis scale and have the labels there instead. I could see there being an issue with long value labels pushing the variable label too far to the left unless we truncate the labels. In the case that there are boxes in the first column that are too small for the label or the color isn't visible, would it be possible to draw a short line from the label to where that box would be, so the labels could still be in the same order as the values, but offset if needed? And if you moused over the label, could it highlight all associated boxes for the user?

![image](https://user-images.githubusercontent.com/54597319/130134090-7e99f662-7da6-4611-a1f9-422cf6e0b30f.png)

chowington:
@SheenaTomko Spent some more time thinking about this---it would certainly be a problem. Consider this mosaic with hover labels:

![image](https://user-images.githubusercontent.com/12377649/131231081-d249d943-f802-4a89-b67e-e9734983c36b.png)

The left-hand y-axis labels you describe are essentially these hover labels but always visible. In the image the tiny boxes at the bottom correspond to a lot of vertical space taken up by labels. You can image this getting even worse, such that the labels take up more vertical space than the whole plot itself. That's fine with hover labels because they're usually hidden, but these axis labels would always be visible. I don't think that's feasible.

I think this is no longer just an implementation problem. It's a UI problem that needs to be sorted out before talking about implementation issues.

d-callan:
[...]

PR #238, opened 2021-10-28
Mosaic left legend

chowington:
Bit of a messy PR due to having to calculate lots of stuff plotly doesn't tell you easily.

Requires Plotly 2.x, so blocked by https://github.com/VEuPathDB/web-components/pull/233

Works toward https://github.com/VEuPathDB/web-components/issues/184

### DIFF
--- a/src/plots/MosaicPlot.tsx
+++ b/src/plots/MosaicPlot.tsx
@@ -10,6 +10,7 @@ import { PlotParams } from 'react-plotly.js';
 import _ from 'lodash';
 // util functions for handling long tick labels with ellipsis
 import { axisTickLableEllipsis } from '../utils/axis-tick-label-ellipsis';
+import { makeStyles } from '@material-ui/core/styles';
 
 export interface MosaicPlotProps extends PlotProps<MosaicData> {
   /** label for independent axis */
@@ -27,6 +28,15 @@ export const EmptyMosaicData: MosaicData = {
   dependentLabels: [],
 };
 
+const useStyles = makeStyles({
+  root: {
+    '& .legend .traces .legendtoggle': {
+      // Remove the click/pointer cursor from legend items
+      cursor: 'default !important',
+    },
+  },
+});
+
 const MosaicPlot = makePlotlyPlotComponent(
   'MosaicPlot',
   ({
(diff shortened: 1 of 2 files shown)


bobular (commented):
Looks great! I have one issue from testing, and a suggestion.
Here I tweaked the dependent labels in Storybook controls and got an overlap
image

Also, if `displayLegend` is false, I think the y-axis title standoff needs to be zero and the margin reduced (so that the plot takes up most of the plotting area).  This would be needed for future faceting implementations, but NOT for b55.  So it's either a do-now or make-a-ticket thing :-)

chowington:
@bobular About the first issue (the storybook one), I wasn't able to reproduce that actually. What did you do to see that, exactly?

chowington:
@bobular Your second comment has been addressed 👍

chowington:
Click/pointer cursor has been fixed

bobular:
> @bobular Thanks for the feedback! About the first issue (the storybook one), I wasn't able to reproduce that actually. When I change the labels using storybook controls, they don't overlap the title.
> 
> What did you do to see that, exactly?

I did the same thing just now on a different machine and it works great.

When I had the problem, it was with an older version of Firefox on Windows.  I'm not going to test it again right now.  Maybe later.  Sorry for any confusion.

bobular (approved):
All looking tip-top now - nice one, thanks!

dmfalke (approved):
Looks great! Nice work.

src/plots/MosaicPlot.tsx
```diff
@@ -30,6 +45,11 @@ const MosaicPlot = makePlotlyPlotComponent(
     dependentAxisLabel,
     colors,
     showColumnLabels,
+    spacingOptions,
+    containerClass,
+    // Pull legendOptions out to ignore it. Mosaic's left legend doesn't really
+    // function as a legend and is less flexible by necessity
+    legendOptions,
```
chowington, line 52: @bobular Note this new change. Is this acceptable rather than making legendOptions an optional addon that every plot except mosaic uses? The only downside to this way is it does still show up in mosaic's API

bobular: Hi @chowington - this is fine with me, rather than introducing addons everywhere else.

Might need something like this to avoid warnings
```
 /* eslint-disable no-unused-vars */
```

chowington: I'm not getting any errors or warnings for it as far as I can tell. Maybe it's fine? I'll go ahead and merge. Thanks Bob!

PR #242, opened 2021-11-02
matching new version of plotly.js with web-eda

moontrip:
When upgrading plotly.js version at web-eda, I realized that they released a new version, v2.6.0, while our upgraded version was v2.5.1. For matching the version between the two repos, web-components and web-eda, I have made this PR. From my tests, all stories at web-components seem to work fine like v2.5.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "leaflet": "^1.6.0",
     "leaflet-drift-marker": "^1.0.3",
     "luxon": "^1.25.0",
-    "plotly.js": "^2.5.1",
+    "plotly.js": "^2.6.0",
     "re-resizable": "^6.9.0",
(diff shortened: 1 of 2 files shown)


moontrip:
I have also looked at plotly.js's release note but no specifically potential issue was noticed for our cases: https://github.com/plotly/plotly.js/releases/

chowington (approved):
I quickly reviewed the stories too, and the looked good. Should be fine since it's a minor version update

bobular (approved):

bobular:
Are you around @moontrip ?  If not, I may merge this + EDA sibling myself.  Because I need to do a "@veupathdb/components" upgrade on my facet branch, and it will be easier to do this first.

PR #248, opened 2021-11-09
Facet container

dmfalke:
This PR includes changes to make the faceted plot container element scrollable. The changes are pretty minimal.

Some decisions are needed on how to best apply plot-specific container styles. The most viable short-term approach is to set the container styles in the consuming component (i.e., in web-eda visualizations). Once we have a better handle on what the ideal plot-specific container styles are, we can consider creating plot-specific `FacetedPlot` components in this repo.

ETA I used the CSS grid layout for the container element. This is future-looking for when we enable wrapping and multi facet variables.

### DIFF
--- a/src/plots/FacetedPlot.tsx
+++ b/src/plots/FacetedPlot.tsx
@@ -1,46 +1,39 @@
 import React from 'react';
-import { isFaceted } from '../types/guards';
 import { FacetedData } from '../types/plots';
 import { PlotProps } from './PlotlyPlot';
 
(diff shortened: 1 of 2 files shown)


bobular (approved):
Looks great in storybook.
It looks like on the EDA side we will have to make a fixed-width container for the scrolling magic to happen, which is of course no biggie.

dmfalke:
Thanks, @bobular. I will make a PR for the web-eda side of this soon.

Issue #241, opened 2021-11-02
Faceting: Thumbnail image capture for faceted plots


bobular:
Subtask of https://github.com/VEuPathDB/web-eda/issues/635

Not sure how much of this is on the EDA-side, but made the ticket here anyway.

PR #250, opened 2021-11-09
Introduce refs for FacetedPlot

jtlong3rd:
(Relates to https://github.com/VEuPathDB/web-components/issues/241.)

This PR introduces refs for `FacetedPlot`s.

A `FacetedPlot` ref is defined to be an _array_ of the `PlotRef`s for each rendered subplot; in particular, a `FacetedPlot` instance exposes each subplot's `toImage` callback, so that we can create assemble thumbnails from multiple subplots.

**Future work:**

@dmfalke recently suggested that we make plot-specific `FacetedPlot` components (e.g., `FacetedHistogram`) and apply plot-specific styles there; when it comes time to do that work, the `makeFacetedPlotComponent` factory which I introduced here might help.

### DIFF
--- a/src/plots/FacetedPlot.tsx
+++ b/src/plots/FacetedPlot.tsx
@@ -1,20 +1,37 @@
-import React from 'react';
+import React, { Ref, forwardRef, useImperativeHandle, useRef } from 'react';
+
+import { memoize } from 'lodash';
+
 import { isFaceted } from '../types/guards';
 import { FacetedData } from '../types/plots';
-import { PlotProps } from './PlotlyPlot';
+import { ComponentWithPlotRef, PlotProps, PlotRef } from './PlotlyPlot';
 
 export interface FacetedPlotProps<D, P extends PlotProps<D>> {
   data?: FacetedData<D>;
-  component: React.ElementType<P>;
+  component: ComponentWithPlotRef<P>;
   props: P;
 }
 
-export default function FacetedPlot<D, P extends PlotProps<D>>(
-  props: FacetedPlotProps<D, P>
-) {
-  const { data, component, props: componentProps } = props;
+export type FacetedPlotRef = PlotRef[];
 
-  const Component = component as React.ElementType; // casting seems to be needed if using component: React.ElementType<P>; above
+function renderFacetedPlot<D, P extends PlotProps<D>>(
+  props: FacetedPlotProps<D, P>,
+  ref: Ref<FacetedPlotRef>
+) {
+  const { data, component: Component, props: componentProps } = props;
+  const plotRefs = useRef<FacetedPlotRef>([]);
+
+  useImperativeHandle<FacetedPlotRef, FacetedPlotRef>(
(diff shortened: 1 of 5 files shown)


src/plots/FacetedPlot.tsx
```diff
@@ -26,6 +56,13 @@ export default function FacetedPlot<D, P extends PlotProps<D>>(
         {data?.facets.map(({ data, label }, index) => (
           <Component
             {...componentProps}
+            ref={(plotInstance) => {
```
bobular, line 59: Is this function a bit like the `(prev) => prev+1` functions you can pass to `setState` functions returned by `useState`?

So it will give us a way to access the ref already made for the individual plot component?

bobular (approved):
As far as I can wrap my head around this - looks good!
Now I'll take a look on the EDA side.

dmfalke (approved):

src/plots/FacetedPlot.tsx
jtlong3rd: Yup, that's the basic idea - the [ref callback](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) "exposes" the`Component`'s underlying ref/instance. More precisely, the callback is invoked when the `Component` is mounted (in which case the underlying `PlotRef` is passed to the callback) or unmounted (in which case `null` is passed to the callback).

When the `index`-th `Component` is mounted or unmounted, the ref callback executes the following logic:

1. If `plotInstance` is nil (that is, the `Component` is being unmounted), delete the `index`-th entry of `plotRefs.current`.
2. Otherwise, set the `index`-th entry of `plotRefs.current` to the newly-mounted `plotInstance`.

PS: If `FacetedPlot` were rendering a fixed number of `Components`, I could have just maintainted a `plotRef` for each `Component` instance in separate `useRef`s, instead of using a ref callback.

```typescript
function FacetedPlot(props) {
  const plotRef1 = useRef<PlotRef>(null);
  const plotRef2 = useRef<PlotRef>(null);
  const plotRef3 = useRef<PlotRef>(null);
  const plotRef4 = useRef<PlotRef>(null);

  useImperativeHandle<FacetedPlotRef, FacetedPlotRef>(
    () => [
      plotRef1.current,
      plotRef2.current,
      plotRef3.current,
      plotRef4.current,
    ]
  );

  return (
    <>
      {/* some unrelated TSX */}
      <Component ref={plotRef1} {/* some props for plot 1 */} />
      <Component ref={plotRef2} {/* some props for plot 2 */} />
      <Component ref={plotRef3} {/* some props for plot 3 */} />
      <Component ref={plotRef4} {/* some props for plot 4 */} />
    </>
  );
);
```

Because the number of rendered `Component`s is in fact dynamic, and the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) require the number of hook invocations to be constant between renders, maintaining each individual `Component`'s ref in separate `useRef`s wasn't an option for me.

PR #251, opened 2021-11-10
Custom legend markers and colors

moontrip:
This PR will address custom legend markers, colors, etc. for corresponding Vizs.

### DIFF
--- a/src/components/plotControls/PlotLegend.tsx
+++ b/src/components/plotControls/PlotLegend.tsx
@@ -45,9 +45,9 @@ export default function PlotLegend({
   };
 
   //DKDK set marker sizes
-  const defaultMarkerSize = '1.2em';
+  const defaultMarkerSize = '1.0em';
   const circleMarkerSize = '1.0em';
-  const markerBorderWidth = '0.2em';
+  const markerBorderWidth = '0.125em';
 
   return (
     <>
--- a/src/components/plotControls/PlotLegend.tsx
+++ b/src/components/plotControls/PlotLegend.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Checkbox } from '@material-ui/core';
+import * as ColorMath from 'color-math';
 
 // define legendItems props
 export interface LegendItemsProps {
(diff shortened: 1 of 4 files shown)


moontrip:
now it is ready for Histogram, Barplot, and Boxplot Vizs.

jtlong3rd (approved):

src/components/plotControls/PlotLegend.tsx
```diff
@@ -103,6 +110,29 @@ export default function PlotLegend({
...
+                          borderColor:
+                            checkedLegendItems?.includes(item.label) &&
+                            item.hasData
+                              ? item.markerColor
+                              : '#999',
+                          backgroundColor:
+                            checkedLegendItems?.includes(item.label) &&
+                            item.hasData
+                              ? item.markerColor
+                                  ?.substring(0, item.markerColor?.length - 1)
```
bobular, line 129: I'm a bit worried that we do not enforce `rgb(int,int,int)` values for markerColor - so if a hex color is provided, this will be unpredictable.

Might be better to use this
https://www.npmjs.com/package/color-math 

```node
var colorMath = require("color-math")
var result = colorMath.evaluate("rgb(124,22,230) @a 30%")
console.log(result.result.css())
// prints rgba(124,22,230,0.3)
```

This will safely handle all kinds of colours, including words like "red"

bobular (approved):
Approving - please add the color-math as a technical debt ticket (and add comments to the markerColor prop that it requires rgb(r,g,b) format) if you don't want to add this now.

src/components/plotControls/PlotLegend.tsx
moontrip: @bobular Thank you for your review and suggestion. Yes I agree with your opinion to safely handle colors. So, as suggested I introduced color-map and changed the code accordingly.

moontrip:
This PR addresses a part of the ticket, [Visualization: Custom legend markers and colors #678](https://github.com/VEuPathDB/web-eda/issues/678): Boxplot Viz is done in this PR

PR #258, opened 2021-11-16
Use vertical ticks for mosaic plot

### LINKED ISSUE #257: Mosaic plot: Make ticks vertical for consistent plot dimensions

dmfalke:
When x-axis ticks are displayed diagonally, it can cause the width of the plot to vary, since plotly factors the amount of space needed by ticks as a part of the total plot dimensions.

**Before**
image

**After**
image

closes #257


bobular:
Possibly an edge case (no pun intended) with single digit number variables?

(WASHb Bangladesh)
image

dmfalke:
Thanks. I assumed we want this for non-faceted plots so they always use the
same horizontal space. We can see what others think.

bobular:
But non-faceted plots don't need to be so strictly equal-widthed.

PR #262, opened 2021-11-16
Faceting title and empty data

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/681

Sorry, I got a little lost in the adjacent work that was done @bobular. Does this reflect what we've discussed surrounding the above ticket?

### DIFF
--- a/src/plots/FacetedPlot.tsx
+++ b/src/plots/FacetedPlot.tsx
@@ -77,6 +77,7 @@ function renderFacetedPlot<D, P extends PlotProps<D>>(
             interactive={false}
             // pass checkedLegendItems to PlotlyPlot
             checkedLegendItems={checkedLegendItems}
+            showNoDataOverlay={data == null}
           />
         ))}
       </div>
--- a/src/plots/FacetedPlot.tsx
+++ b/src/plots/FacetedPlot.tsx
@@ -72,7 +72,7 @@ function renderFacetedPlot<D, P extends PlotProps<D>>(
             }}
             key={index}
             data={data}
-            title={data == null ? `NO DATA: ${label}` : label}
+            title={label}
             displayLegend={false}
             interactive={false}
             // pass checkedLegendItems to PlotlyPlot
--- a/src/plots/MosaicPlot.tsx
+++ b/src/plots/MosaicPlot.tsx
@@ -11,6 +11,7 @@ import _ from 'lodash';
 // util functions for handling long tick labels with ellipsis
 import { axisTickLableEllipsis } from '../utils/axis-tick-label-ellipsis';
 import { makeStyles } from '@material-ui/core/styles';
+import { PlotSpacingDefault } from '../types/plots/addOns';
 
 export interface MosaicPlotProps extends PlotProps<MosaicData> {
   /** label for independent axis */
@@ -87,10 +88,6 @@ const MosaicPlot = makePlotlyPlotComponent(
     // The gap between each legend item
     let legendTraceGroupGap: number | undefined;
 
-    // ploltly.js default for left, bottom, and right margins is the same
-    const defaultMargin = 80;
-    const defaultMarginTop = 100;
-
     if (restProps.displayLegend) {
       // Not currently calculating this---just using the default
       // Might need to be calculated or adjusted if more flexibility is needed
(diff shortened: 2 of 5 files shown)


bobular (approved):
Cool - looks great.  Please could you just add the following to FacetedPlot (`<Component ... >`)
```
  showNoData={data == null}
```
Then we can test it out in the EDA (all merged now).
I have a feeling the dark gray will be a bit overpowering.  But we can tweak that to our hearts content later on.

bobular:
Oh wait, I found something not quite right.

I played with the Boxplot story controls, and added my own title 

image

I'm pretty sure the title shouldn't be obscured - so more like this that I tweaked in dev tools:

image
(though ignore the large gray margin at the bottom - it is still 100% height I think)

bobular:
Thanks for the FacetedPlot change.

Here's how it looks in the EDA.  Possibly a bit in your face?
image

bobular:
We can get rid of the "NO DATA:" prefix now too :smile:

chowington:
Thanks for the feedback @bobular! I'll work on those changes

chowington:
@bobular Check it out now!

chowington:
@bobular What was the conclusion of the EDA dev meeting on this? Any styling changes?

bobular:
> @bobular What was the conclusion of the EDA dev meeting on this? Any styling changes?

Hi - yes there were.  Lighter grey background (and maybe even higher opacity to achieve the same level of obscuration, if that's a word), smaller "No data" font size and try white text for "No data".  Thanks!

bobular:
I just found an edge case with scatter plot that doesn't look so good.

Could we gray out the entire plot (like you did at first!), have the `No data` text in #e8e8e8 gray (this is 'EDA no data gray') and put the plot title on top of everything, in black?

image

chowington:
`git revert` ;)

chowington:
@bobular Did you envision a different color for the overlay background? Setting the text to EDA No Data Gray© makes the text hard to see against the current background color, however making the background darker makes it more prominent on the page, which seemed to be an issue in the meeting.

bobular:
I played around with grays. It seems we can go *really* light and opaque and it looks OK.

              backgroundColor: '#f8f8f8f8',
	      color: '#e8e8e8',

bobular:
Danica would like some special formatting for the title of the "No data" facet (that's not to be confused with facets with no data!!)

So when `title === 'No data'` some subtly distinctive styling.  Maybe including special no data gray #e8e8e8. Perhaps a border or background?

chowington:
> 
> 
> Danica would like some special formatting for the title of the "No data" facet (that's not to be confused with facets with no data!!)
> 
> So when `title === 'No data'` some subtly distinctive styling. Maybe including special no data gray #e8e8e8. Perhaps a border or background?

Hmmm, now it seems like we're really getting into `web-eda` concerns that don't belong in `web-components`... This is along the same lines when I was worried about building the overlay into `web-components` when we're not sure the overlay will be useful for every library that uses it. That's why I first suggested having `web-eda` create the overlay itself if it wants one and not have `web-components` worry about it at all. Thoughts?

bobular:
> Hmmm, now it seems like we're really getting into `web-eda` concerns that don't belong in `web-components`... This is along the same lines when I was worried about building the overlay into `web-components` when we're not sure the overlay will be useful for every library that uses it. That's why I first suggested having `web-eda` create the overlay itself if it wants one and not have `web-components` worry about it at all. Thoughts?

Absolutely agree there is EDA-creep going on here.  And @dmfalke made a similar suggestion [here](https://github.com/VEuPathDB/web-eda/issues/706#issuecomment-969843766).  We already have some instances of `'No data'` checks in `web-components` - so this wouldn't be the first time.  Two wrongs don't make a right but we do have to get something working by tomorrow!  I suggest being pragmatic and we can make a technical-debt ticket as needed.

chowington:
@bobular I just pushed the following. Let me know what you think:

image

bobular:
The "No missing data" is a nice addition - love the stripes subtly drawing attention to something different going on here - though the need for it might go away after https://github.com/VEuPathDB/web-eda/issues/676 (not scheduled for b55).

What Danica envisaged was something to make the 'No data' facet (when there **is** data for it, as in the screenshot below - yeah this is confusing even for us!) look distinctive.  I'll have a quick play with some simple ideas and post some images
image

bobular (approved):
Yeah I think just go with Italics to match the legend.  There was a suggestion to use the No Data Gray :tm:  but I think that might be confusing (people will go hunting around for gray maybe)

I'll push the tiny change after this.

Issue #265, opened 2021-11-17
Mosaic: Y-axis labels container sometimes overflows plot height


chowington:
The mosaic y-axis labels (it's not a legend *shhh*) container overflows the height of the plot and gets a scrollbar if the x-axis labels are too long. This is because the y-axis labels container height is calculated based on the assumed minimum height of the plot, which has become smaller now that the x-axis labels are vertical rather than slanted.

chowington:
See this comment for a test case: https://github.com/VEuPathDB/web-components/pull/266#pullrequestreview-809860627

chowington:
Image of y-axis labels overflowing their container---notice scroll bar to their right. This is on GEMS1.

![image](https://user-images.githubusercontent.com/12377649/142512847-57a66050-136b-467f-9ab5-695ceba5c5e9.png)

aaronwlsong:
It looks good now, closing ticket.

In GEMS1:

![image](https://user-images.githubusercontent.com/61077845/142640546-d9f6d5d5-ace4-48f1-818a-40975a1c861f.png)

In other studies: 
![image](https://user-images.githubusercontent.com/61077845/142640958-ff9b3ebc-a9c0-4283-8e89-16e036abe0c5.png)

PR #266, opened 2021-11-17
Fix bug that allowed the y-axis labels to overflow the plot vertically

chowington:
Works toward https://github.com/VEuPathDB/web-components/issues/265

### DIFF
--- a/src/plots/MosaicPlot.tsx
+++ b/src/plots/MosaicPlot.tsx
@@ -38,6 +38,15 @@ const useStyles = makeStyles({
   },
 });
 
+const getLongestTruncatedStringLength = (
+  strings: string[],
+  maxLength: number
+) => {
+  const longestStringLength = Math.max(...strings.map((str) => str.length));
+  // If the length overflows, add two characters to account for ellipsis length
+  return longestStringLength > maxLength ? maxLength + 2 : longestStringLength;
+};
+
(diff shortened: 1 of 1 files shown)


src/plots/MosaicPlot.tsx
```diff
@@ -118,12 +127,16 @@ const MosaicPlot = makePlotlyPlotComponent(
...
+          data.independentLabels,
+          maxIndependentTickLabelLength
+        );
         // Subtraction at end is due to x-axis automargin shrinking the plot
         const plotHeight =
           containerHeight -
           marginTop -
           marginBottom -
-          2.05 * (maxIndependentTickLabelLength + 2);
+          5 * longestIndependentTickLabelLength;
```
chowington, line 139: This is really the main fix, changing from 2px per character to 5px to account for vertical strings taking up more space than slanted ones.

The rest is just for greater flexibility, basing the longest label length on the lengths of the labels actually provided rather than simply the maximum label length in general.

bobular (approved):
This looks good.  I can't remember what conditions triggered the scrolling y-axis labels "legend" but it's not happening now.

bobular:
Just a reminder that we'll need to do an EDA components upgrade at some point. Could be done at the same time as `faceting-title-and-empty-data` perhaps.

PR #270, opened 2021-11-19
use arrow mouse pointer for custom legend title and labels

### LINKED ISSUE #271: New custom legend mouse pointer feedback is confusing
Tested with FF in Linux

When mousing over the text in the legend, the mouse pointer becomes a finger, suggesting clicking will do something, but it doesn't.

Pointer behaviour when mousing-over the coloured square, circle or lines is as expected
Same for the checkboxes.

But the text is disconcerting.

Perhaps the text should be clickable too (but not the legend title).

And maybe double click should emulate what Plotly does?  (very optional - get feedback from outreach first!)

moontrip:
Resolves #271 

Going to pause the issue and put this in draft mode.  Needs further discussion with outreach.

### DIFF
--- a/src/components/plotControls/PlotLegend.tsx
+++ b/src/components/plotControls/PlotLegend.tsx
@@ -77,6 +77,7 @@ export default function PlotLegend({
                     id={item.label}
                     value={item.label}
                     color="primary"
+                    // color="default"
                     onChange={(e) => {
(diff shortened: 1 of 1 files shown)


moontrip:
This is not what we will do. Rather we will support clickable legend label, so the mouse pointer would be just fine

Issue #259, opened 2021-11-16
Faceting: individual `FacetedPlot` types


dmfalke:
This is a follow-up to #240, which lists tasks for `FacetedPlot`.

- [x] make `FacetedPlot` components for each plot type (e.g., `FacetedBarPlot`)
- [x] make simple story for each plot type (PiePlot is done)
- [x] click-on-plot to an enlarged modal
~~"No data" facet styled differently (perhaps the title has a grey background)~~ Fixed in https://github.com/VEuPathDB/web-components/pull/262

chowington:
Striking out the third task above because it was done in https://github.com/VEuPathDB/web-components/pull/262

dmfalke:
@chowington I must have authored this issue in haste. It is missing a crucial component, which is to make a `FacetedPlot` component for each of the plot types we provide. Ideally, they should use styling similar to what we are doing in visualizations in web-eda. I will update the description to reflect this. Sorry for the confusion. The stories do look great!

chowington:
@dmfalke I started work on the individual plot components but I don't have a great idea of how to integrate them into our existing structure, particularly with `FacetedPlot`. I just worked on the `FacetedBarplot` as a test run and pushed it to a different branch than this one: https://github.com/VEuPathDB/web-components/tree/individual-facetedplot-components. It's not currently functional, but you can see what I'm attempting to do. Let me know what your thoughts are.

chowington:
Bumping up to high cost due to extra core-components modal work for good housekeeping

nkittur-uga:
Dave advised that the primary ux change is the ability to click on a facet to see an enlarged plot. I tested this and it works. Passes QA.

PR #272, opened 2021-11-22
Faceting #259

chowington:
Works toward https://github.com/VEuPathDB/web-components/issues/259

### DIFF
--- a/src/plots/FacetedPlot.tsx
+++ b/src/plots/FacetedPlot.tsx
@@ -23,8 +23,8 @@ type ComponentWithPlotRef<P> = ComponentType<
 export interface FacetedPlotProps<D, P extends PlotProps<D>> {
   data?: FacetedData<D>;
   component: ComponentWithPlotRef<P>;
-  props: P;
-  modalProps?: P;
+  componentProps: P;
+  modalComponentProps?: P;
   // custom legend prop
   checkedLegendItems?: string[];
 }
@@ -36,9 +36,9 @@ function renderFacetedPlot<D, P extends PlotProps<D>>(
   const {
     data,
     component: Component,
-    props: componentProps,
-    modalProps: modalComponentProps,
-    checkedLegendItems: checkedLegendItems,
+    componentProps,
+    modalComponentProps,
+    checkedLegendItems,
   } = props;
   const plotRefs = useRef<FacetedPlotRef>([]);
   const [modalIsOpen, setModalIsOpen] = useState(false);
@@ -68,59 +68,46 @@ function renderFacetedPlot<D, P extends PlotProps<D>>(
         }}
       >
         {data?.facets.map(({ data, label }, index) => {
-          const component = (
-            <Component
-              {...componentProps}
-              ref={(plotInstance) => {
-                if (plotInstance == null) {
-                  delete plotRefs.current[index];
-                } else {
-                  plotRefs.current[index] = plotInstance;
-                }
-              }}
-              key={index}
-              data={data}
-              title={label}
-              displayLegend={false}
-              interactive={false}
-              // pass checkedLegendItems to PlotlyPlot
-              checkedLegendItems={checkedLegendItems}
-              showNoDataOverlay={data == null}
-            />
-          );
+          const sharedProps = {
+            data: data,
+            title: label,
+            // pass checkedLegendItems to PlotlyPlot
+            checkedLegendItems: checkedLegendItems,
+            showNoDataOverlay: data == null,
+          };
 
-          const modalComponent = modalComponentProps && (
-            <Component
-              {...modalComponentProps}
-              // ref={(plotInstance) => {
-              //   if (plotInstance == null) {
-              //     delete plotRefs.current[index];
-              //   } else {
-              //     plotRefs.current[index] = plotInstance;
-              //   }
-              // }}
-              // key={index}
-              data={data}
-              title={label}
-              displayLegend={true}
-              interactive={true}
-              // pass checkedLegendItems to PlotlyPlot
-              checkedLegendItems={checkedLegendItems}
-              showNoDataOverlay={data == null}
-            />
-          );
+          const divProps = modalComponentProps && {
+            onClick: () => {
+              setModalPlot(
+                <Component
+                  {...modalComponentProps}
+                  {...sharedProps}
+                  displayLegend={true}
+                  interactive={true}
+                />
+              );
+              setModalIsOpen(true);
+            },
+            style: { cursor: 'pointer' },
+            title: 'Click to expand',
+          };
 
-          return modalComponentProps ? (
-            <button
-              onClick={() => {
-                setModalPlot(modalComponent);
-                setModalIsOpen(true);
-              }}
-            >
-              {component}
-            </button>
-          ) : (
-            <>{component}</>
+          return (
+            <div {...divProps} key={index}>
+              <Component
+                {...componentProps}
+                {...sharedProps}
+                ref={(plotInstance) => {
+                  if (plotInstance == null) {
+                    delete plotRefs.current[index];
+                  } else {
+                    plotRefs.current[index] = plotInstance;
+                  }
+                }}
+                displayLegend={false}
+                interactive={false}
+              />
+            </div>
           );
         })}
       </div>
--- a/src/plots/FacetedPlot.tsx
+++ b/src/plots/FacetedPlot.tsx
@@ -112,11 +112,20 @@ function renderFacetedPlot<D, P extends PlotProps<D>>(
         })}
       </div>
       {modalComponentProps && (
-        <FullScreenModal
-          visible={modalIsOpen}
-          // onClose={() => setModalIsOpen(false)}
-        >
-          <button onClick={() => setModalIsOpen(false)}>Close</button>
+        <FullScreenModal visible={modalIsOpen}>
+          <button
+            onClick={() => setModalIsOpen(false)}
+            style={{
+              position: 'absolute',
+              top: 30,
+              right: 30,
+              backgroundColor: 'white',
+              cursor: 'pointer',
+              border: 'none',
+            }}
+          >
+            <i className="fas fa-times fa-lg"></i>
+          </button>
           {modalPlot}
         </FullScreenModal>
       )}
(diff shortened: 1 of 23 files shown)


dmfalke:
@chowington I updated the linked issue with some missing details. It may impact the work you've done here.

bobular (approved):
Hi!
Looks to me like there's a bug in the Storybook Controls UI

In Barplot, for example, I would say there should only be 2 controls: `data` and `props`, but it lists 27.  25 of them are inoperative.  But I guess we shouldn't dwell on this right now.

Looks great otherwise, cheers.

bobular:
> Looks to me like there's a bug in the Storybook Controls UI
> 
> In Barplot, for example, I would say there should only be 2 controls: `data` and `props`, but it lists 27. 25 of them are inoperative. But I guess we shouldn't dwell on this right now.

Maybe the controls are shared per story tsx file?  Or leak between stories in the same file, or something?  Not worth investigating further, just thought I'd make a note of this here.

dmfalke:
I took a very quick look at the code, and I think you're going in the right direction. I haven't tested anything, but hope to do so in the next day or two.

chowington:
@bobular @dmfalke The main tasks for this PR's ticket are mostly done. Do take a look please! There's one outstanding bug and potentially other improvement to make that we can discuss.

bobular (approved):
This looks great. I tested also by nuking `modalComponentProps` in storybook controls and it works great with the modals disabled.

Am I correct in thinking that if we have N facets, we have N on-screen components that each "contain" data.  We also have N lots of divModalProps.onClick handlers which also contain a copy of the data (in `sharedProps`).  Is that using twice the memory? Or is it just a copy by reference.  And even if it is a deep copy, does it even matter?

Excellent work with the `interactive`/`displayLegend` stuff and I think it is essential to set the modal plot's `title`, so all :+1: from me.

bobular:
Oh there is one small thing.  If you check the Histogram story, you see that the modal plot has a legend with a single item.

In EDA we disable the legend when there's only one data series.  So we will want to be able to control `displayLegend` from the upstream component, rather than have it always `true` as below

```
          const divModalProps = modalComponentProps && {
            onClick: () => {
              setModalPlot(
                <Component
                  {...sharedProps}
                  displayLegend={true}
                  interactive={true}
                  {...modalComponentProps}
                  title={label}
                />
              );
              setModalIsOpen(true);
            },
            style: { cursor: 'pointer' },
            title: 'Click to expand',
          };
```

chowington:
> This looks great. I tested also by nuking `modalComponentProps` in storybook controls and it works great with the modals disabled.
> 
> Am I correct in thinking that if we have N facets, we have N on-screen components that each "contain" data. We also have N lots of divModalProps.onClick handlers which also contain a copy of the data (in `sharedProps`). Is that using twice the memory? Or is it just a copy by reference. And even if it is a deep copy, does it even matter? (thinking scatterplots)
> 
> Excellent work with the `interactive`/`displayLegend` stuff and I think it is essential to set the modal plot's `title`, so all 👍 from me.

Thanks Bob!

I believe JS's object passing behavior is a (slightly-modified) form of pass-by-reference, so theoretically we aren't duplicating the data here. Definitely a valid concern though!

chowington:
> Oh there is one small thing. If you check the Histogram story, you see that the modal plot has a legend with a single item.
> 
> In EDA we disable the legend when there's only one data series. So we will want to be able to control `displayLegend` from the upstream component, rather than have it always `true` as below
> 
> ```
>           const divModalProps = modalComponentProps && {
>             onClick: () => {
>               setModalPlot(
>                 <Component
>                   {...sharedProps}
>                   displayLegend={true}
>                   interactive={true}
>                   {...modalComponentProps}
>                   title={label}
>                 />
>               );
>               setModalIsOpen(true);
>             },
>             style: { cursor: 'pointer' },
>             title: 'Click to expand',
>           };
> ```

This should already allow that! The current logic allows you to override `displayLegend={true}` by passing your own `displayLegend` value in `modalComponentProps`, as `{...modalComponentProps}` overwrites the previous props 👍

bobular:
Oh yeah, sorry I didn't spot that!

chowington:
> @bobular @dmfalke The main tasks for this PR's ticket are mostly done. Do take a look please! There's one outstanding bug and potentially other improvement to make that we can discuss.

Sorry, I was a little vague late last night! More details:

- Bug: Clicking directly on the pie chart portion of a pie facet doesn't trigger the click handler that opens the modal. You have to click outside the pie itself. I'm not sure why exactly this is happening yet; if anyone has an idea, let me know!
- Comment: I'll probably double check the layout prop defaults to make sure they are reasonable
- Question: Should we have defaults for the modal component layout props as well?
- Question: How should we handle the mosaic contingency table? Should there be a modal for those tables as well? I didn't test that at all and there isn't a story for the cont table, faceted or otherwise

chowington:
@dmfalke @bobular Any comments on the above?

bobular:
Thanks @chowington - having a look now.

Unrelated to your questions, I saw this in the console, FYI:

> error: "Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}`."

dmfalke:
> > @bobular @dmfalke The main tasks for this PR's ticket are mostly done. Do take a look please! There's one outstanding bug and potentially other improvement to make that we can discuss.
> 
> Sorry, I was a little vague late last night! More details:
> 
>     * [x]  Bug (Fixed): Clicking directly on the pie chart portion of a pie facet doesn't trigger the click handler that opens the modal. You have to click outside the pie itself. I'm not sure why exactly this is happening yet; if anyone has an idea, let me know!
> 
>     * [ ]  Comment: I'll probably double check the layout prop defaults to make sure they are reasonable
> 
>     * [ ]  Question: Should we have defaults for the modal component layout props as well?
I think it makes sense to have defaults.

> 
>     * [ ]  Question: How should we handle the mosaic contingency table? Should there be a modal for those tables as well? I didn't test that at all and there isn't a story for the cont table, faceted or otherwise
My initial thought is to have the same set of tabs we include for an unfaceted mosaic plot.

src/plots/facetedPlots/FacetedPiePlot.tsx
```diff
@@ -0,0 +1,59 @@
...
+type FacetedPiePlotProps = Omit<
+  FacetedPlotProps<PiePlotData, PiePlotProps>,
+  'component'
+>;
+
+const FacetedPiePlot = (facetedPiePlotProps: FacetedPiePlotProps) => {
+  const onPlotlyRender = useCallback((_, graphDiv) => {
+    // Replace each slice DOM node with a copy of itself.
+    // This removes the existing click event handler, which
+    // otherwise blocks the click handler that opens the modal
```
bobular, line 31: Thanks for the comment - was looking at an older commit and was going to ask for one!

bobular (approved):
Code looks good, works great.

It would be nice to have the mosaic modal have the table tabs too, but isn't this going to be a nightmare to implement?  Instead of passing `modalComponentProps` to `FacetedMosaicPlot` we'll have to pass an actual component.  Maybe that's an OK hack?

dmfalke:
> Code looks good, works great.
> 
> It would be nice to have the mosaic modal have the table tabs too, but isn't this going to be a nightmare to implement? Instead of passing `modalComponentProps` to `FacetedMosaicPlot` we'll have to pass an actual component. Maybe that's an OK hack?

You're right, I didn't think about that. Let's just stick with the plot for now and see what others think.

dmfalke (approved):
I think this all looks great. One suggestion is to export the default styling and spacing options so that consumers can use them if passing in custom styling or spacing.

chowington:
> Thanks @chowington - having a look now.
> 
> Unrelated to your questions, I saw this in the console, FYI:
> 
> > error: "Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}`."

Yeah, I saw that. The modal comes from `core-components`, and Mike said he'll address that console warning at some point 👍

chowington:
> > * [ ]  Question: Should we have defaults for the modal component layout props as well?
> 
> I think it makes sense to have defaults.

I thought about and worked on this a little bit and I think the resulting API would be a bit odd. Currently `modalComponentProps` existing is the way we activate the click-to-expand feature. If we provide sensible defaults, it would be odd for the user to still have to provide `modalComponentProps` in order to activate the click-to-expand feature.

Instead I attempted to create a better API by providing a `clickToExpand?: boolean` prop that controls the expand feature's activation, however after changing the logic appropriately, there was a complicated TS error that cropped up involving the `P` generic type. I think I understood it, but I didn't see a way to get around it.

So what all this means is that I decided to stick with what we have currently, including not giving default `modalComponentsProps`. How does this sound @dmfalke?

chowington:
> Code looks good, works great.
> 
> It would be nice to have the mosaic modal have the table tabs too, but isn't this going to be a nightmare to implement? Instead of passing `modalComponentProps` to `FacetedMosaicPlot` we'll have to pass an actual component. Maybe that's an OK hack?

Alternatively, we could have the click-to-expand feature for the contingency tables as well, separate from the faceted mosaics. Would that be useful? I forgot what the contingency tables look like in faceted mode.

I am itching to get this merged though, because if another PR gets merge beforehand it could be a nasty merge conflict :)

chowington:
I'm getting ready to wrap this up on the `web-components` side. Just wanted to get one more review from @dmfalke on this and I'll merge: https://github.com/VEuPathDB/web-components/pull/272/commits/ac3926d128d9cfd860506ccbb8b5e25f00e9da4a

dmfalke (approved):
Looks good!

src/plots/FacetedPlot.tsx
```diff
@@ -6,36 +6,51 @@ import React, {
...
+   */
+  modalComponentProps?: P;
   // custom legend prop
   checkedLegendItems?: string[];
 }
 
+export interface FacetedPlotPropsWithRef<D, P extends PlotProps<D>>
+  extends FacetedPlotProps<D, P> {
+  facetedPlotRef?: Ref<FacetedPlotRef>;
+}
```
dmfalke, line 38: :+1:

chowington:
Thanks Dave!

PR #274, opened 2021-11-23
add R-square prop to Scatter plot

moontrip:
While making a R-square table for Scatter plot with the plot options of Best fit with raw ([#694 at web-eda](https://github.com/VEuPathDB/web-eda/issues/694)), it was necessary to add a prop to have R-square value.


jtlong3rd (approved):

moontrip:
Thank you for your quick review as usual 👍 I think it is not harmful to merge this as the new prop is optional so I will merge this and add new version at the corresponding web-eda PR. By doing this, it will be much easier to test the web-eda PR without replying on yarn link

PR #275, opened 2021-12-03
add gradient colormap for scatterplot

asizemore:
Partially addresses web-eda #306.

The goal is to allow for scatter plot points to be colored based on a gradient colorscale. This requires an update to the `processInputData` function, the addition of new colorscales themselves, and a new legend.

The biggest trouble was creating the legend. Ended up doing it mostly by hand, because I had some trouble getting the d3 functions to work as expected. Very interested to hear opinions on how this can be improved! 

Another note on the legend - the organization is sort of funky right now because I am guessing that eventually we'd want it to be part of the main custom legend `PlotLegend`? So for now i've set it in a parent component that mirrors `PlotLegend` but intended to make the actual gradient colormap component so that it could get plopped into `PlotLegend` in the future.


moontrip:
@asizemore interesting work 👍 I am off from VEuPathDB today so I will take a look at this next week!

moontrip:
@asizemore Honestly I did not check all codes and notice a couple of things only when glancing at. It is mainly because I have a couple of questions regarding this gradient legend, before looking at more carefully. 

a) is this gradient legend only for Raw data (markers) ? (not smoothed mean and best fit options)
b) not quite sure if this gradient legend will be used with multiple data series ? One obvious example is with the presence of an overlay variable. I could only see two types of gradient color maps, diverging and sequential gradients at addOns.ts, though.
c) do we have a plan to show both custom legend and gradient legend ?
d) if multiple data are supported with this legend, it is not clear how to distinguish marker colors in custom legend
e) if this only targets Raw data (single dataset), then perhaps we should make an additional plot option like 'Raw with gradient color' which only shows this gradient legend and hide custom legend. Alternatively, a toggle switch may work as well.

asizemore:
So this would be if we selected a continuous overlay variable that has values all >0 (so it's given the sequential colorscale).

Screen Shot 2021-12-08 at 3 49 37 PM

and then perhaps for no data toggle = on it would look like

Screen Shot 2021-12-08 at 3 53 28 PM

?

dmfalke (commented):
This is looking very nice. I'm not seeing a clear case for showing both gradient and checkbox legends. Even with **No data = on**, it would be odd to have both the toggle for **No data** and a checkbox for it. I suppose we might want to reuse some of the rendering logic used by the checkbox legend, which might argue for combining these into one component. At worst, if we do this and determine we need both a gradient and a checkbox legend at the same time, this "super" component can be used twice in the different modes.

moontrip:
Thanks @asizemore for your detained explanation and example screenshot. If I understand correctly, the gradient legend is only for overlay variable with 'continuous' type. Then perhaps we can check dataShape from metadata and show either custom legend or gradient legend selectively.

asizemore:
Alright - all ready for review! 
@moontrip good idea about using dataShape=Continuous. I agree we should do that or just look for dataType=Number. Looking at the code in plot.data and elsewhere, i think we don't have a good idea of what we want the var constraints to be just yet, so i'll bring it up for discussion in one of the eda meetings. 

For more details, see web-eda [PR #790](https://github.com/VEuPathDB/web-eda/pull/790).

asizemore:
As discussed in today's meeting, this PR is paused until after b56.

asizemore:
Hey @dmfalke and @moontrip! 
I was originally planning to address https://github.com/VEuPathDB/web-eda/issues/1041 by only making the changes to `addOns.ts` (adding the two gradient colormaps), since we aren't planning to have these types of gradient color legends nor scatterplots with continuous overlay just yet. However when i revisited this PR, I wondered if it might be helpful to go ahead and merge all of this code since it holds examples of how to work with these colormaps? 

Alternatively i can add the colormaps only (`addOns.ts` changes) and we can leave this PR open for a bit for reference (assuming it's helpful of course). At least this way we don't have code that's not being used yet sitting around in the repository.

I'd be happy either way! Let me know what you two think is best!

asizemore:
Hey @dmfalke and @moontrip! 
I was originally planning to address https://github.com/VEuPathDB/web-eda/issues/1041 by only making the changes to `addOns.ts` (adding the two gradient colormaps), since we aren't planning to have these types of gradient color legends nor scatterplots with continuous overlay just yet. However when i revisited this PR, I wondered if it might be helpful to go ahead and merge all of this code since it holds examples of how to work with these colormaps? 

Alternatively i can add the colormaps only (`addOns.ts` changes) and we can leave this PR open for a bit for reference (assuming it's helpful of course). At least this way we don't have code that's not being used yet sitting around in the repository.

I'd be happy either way! Let me know what you two think is best!

moontrip:
@asizemore my two cents: I think that it can be merged into this repo as they are instructive and will possibly be useful in the future 😃 From my glance, it would not break anything anyway

dmfalke (approved):
Go for it!

PR #281, opened 2021-12-16
Right align table numbers

chowington:
Works toward https://github.com/VEuPathDB/web-eda/issues/761

image

I also right-aligned the column labels that sit above right-aligned numbers. The variable coverage table will be handled in web-eda.

Interestingly, I believe the numbers in this font are equal-width, even though the font's letters are not. I didn't realize some fonts did that. Neat!


chowington:
@bobular Text for headers of numeric columns has been center-aligned as discussed on Slack 👍

bobular (approved):
Nice - thanks!

PR #282, opened 2021-12-17
Map and CSS tweaks to enable use in EDA

bobular:
Progress so far
- commented out the problematic CSS imports.  Need a better solution please anyone!
- named the handlers for bounds (top/bottom/left/right) and viewports (center lat/lon and zoom level) more consistently

### DIFF
--- a/dist/css/map_styles.css
+++ b/dist/css/map_styles.css
@@ -103,3 +103,15 @@
 .plot-marker-popup.popup-down .leaflet-popup-content {
   transform: rotate(180deg);
 }
+
+/* on mouse-over, marker to top */
+.top-marker {
+  z-index: 3100 !important;
+}
+
+/* Replace leaflet styles*/
+
+/* general typography */
+.leaflet-container {
+  font: 12px/1.5 'Open Sans', Arial, Helvetica, sans-serif;
+}
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
   },
   "files": [
     "lib",
-    "src"
+    "src",
+    "dist"
   ],
   "scripts": {
     "start": "yarn storybook",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.11.11",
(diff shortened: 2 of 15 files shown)


bobular:
In the latest commit I moved the most important/functional CSS file into /dist and I think I've set up package.json so that npm will package this.  It seems to work in EDA via yarn link but I am hopeful also for npm...

It seems that these styles have to be handled through static CSS files because the classes are added/removed directly to DOM elements, not React components, though I could be wrong.

bobular:
Dear reviewer... thank you for looking at this PR.
The big question is... will the `dist` packaging of CSS work in npm?
Note that there is a bunch of CSS in `public` still (vb-popbio-maps.css) - this does not get seen by EDA and I don't think anything there is critical for the map component.  We can let vb-popbio-maps.css atrophy or perhaps delete it (or remove it from the storybook header includes?)

dmfalke:
> Dear reviewer... thank you for looking at this PR. The big question is... will the `dist` packaging of CSS work in npm? Note that there is a bunch of CSS in `public` still (vb-popbio-maps.css) - this does not get seen by EDA and I don't think anything there is critical for the map component. We can let vb-popbio-maps.css atrophy or perhaps delete it (or remove it from the storybook header includes?)

If the css files are included in the output folder and included with the npm package, things should work as expected. It is the responsibility of the consumer package to process CSS imports. We do that via webpack in both web-eda and our websites.

dmfalke (approved):

chowington:
@bobular I tested the map on storybook, and when you attempt to zoom in/out, it bounces back to the original zoom level. Seems to work on wed-eda though!

bobular:
> @bobular I tested the map on storybook, and when you attempt to zoom in/out, it bounces back to the original zoom level. Seems to work on wed-eda though!

Ooh, thank you.  I was very bad and didn't test the storybook. Will investigate now.

bobular:
Fixed the stories issue found by Connor.

Going to merge this now and test the CSS stuff via npm

bobular:
Seems to work after yarn unlinking and using npm :-)

PR #283, opened 2021-12-22
Upgrade @emotion/cache so that storybook works again

bobular:
Storybook was failing silently (except from in the console) with an undefined `options`.  Now fixed.


bobular:
Oh, should this have been a dev dependency?

bobular:
Closing in favour of https://github.com/VEuPathDB/web-components/pull/284 being a cleaner fix, if a fix is needed at all.

Issue #264, opened 2021-11-17
When overlay vars are disabled, barplot doesn't keep bars in the same place (like boxplot)


bobular:
Might be EDA-side but probably isn't.

Before and after disabling the 'unknown' dark blue bars:
![image](https://user-images.githubusercontent.com/308639/142249922-d34ce705-cf6f-4809-8b16-66aed677619b.png)
![image](https://user-images.githubusercontent.com/308639/142250003-f3d0affd-3c34-4fc4-964e-871aefbb41e3.png)

Desired behaviour is like already exists in boxplot (bar size and spacing stays the same until only one series remains)

Assuming low cost!

moontrip:
This is Plotly's default behavior (automatic adjustments like changing bar widths) and I am not quite sure if it needs to be changed the way you described

bobular:
Oh, you are right.
https://codepen.io/bobular/pen/WNEPeoR

It's not the same for boxplot though.

We have time to think about this - I didn't put the b55 for a reason!

moontrip:
@bobular yeah somehow our boxplot's behavior is different from Plotly's one. But personally I prefer such a dynamic/automatic behavior because in many cases plots are congested with our limited plot size, especially for Boxplot - e.g., X: Floor material, overlay: Country. We will see~

chowington:
This next cycle ticket has gone a bit stale. Should it go to a UI discussion for which behavior is best? I see merits of both, personally.

bobular:
Desired behaviour would be like barplot (bars resize to use space), so if possible change boxplot behaviour to do this.

aaronwlsong:
I cannot replicate the issue since I dont know what overlay var is in this example. But tested different overly vars, it seems the same situation still happens in barplot. 

eg. GEMS1 Case Control study, when main=age group, overlay=" Primary caregiver's education level", when select and unselect the second one: "Religious education ...", the other bars automatically adjust the spaces. 

<img width="1048" alt="Screen Shot 2022-02-01 at 2 55 35 PM" src="https://user-images.githubusercontent.com/61077845/152041780-bcf3418a-19ce-417b-b352-ad780d30d029.png">

<img width="1089" alt="Screen Shot 2022-02-01 at 2 58 05 PM" src="https://user-images.githubusercontent.com/61077845/152041993-00823414-fad2-4450-aec0-5bc7ee8353a1.png">

PR #288, opened 2022-01-12
Remove boxplot empty gap

chowington:
Works toward https://github.com/VEuPathDB/web-components/issues/264

Thankfully the option I needed to change was commented describing this exact behavior :)

Note that the empty space leftover will disappear even if there is just empty data for the series in question, rather than the series being set to invisible, but that the series' legend entry will remain (this appears to be different than in barplot, where the series legend entry disappears for empty data).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.11.10",
+  "version": "0.11.11",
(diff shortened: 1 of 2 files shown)


bobular (approved):
Yay for comments!

I have tested in EDA and it works great.  Thanks.

PR #289, opened 2022-01-13
Lineplot component

d-callan:
got frustrated w xyplot while working on adding features etc to the lineplot viz in veupathdb/web-eda#645 ..


src/types/plots/lineplot.ts
```diff
@@ -0,0 +1,19 @@
+import { ScatterPlotDataSeries } from '.';
+import { AvailableUnitsAddon } from './addOns';
+import { Bin, BinWidthSlider } from '../general';
+
+type Override<T1, T2> = Omit<T1, keyof T2> & T2;
+
+export type LinePlotDataSeries = Override<
+  ScatterPlotDataSeries,
```
bobular, line 8: Nice Override shortcut!

We're still using a very Plotly-ish data structure for these plots. I think (and am happy to help with this) the ScatterPlotData type should be similar to the back end response type and all the work currently done in ScatterplotVisualization and LineplotVisualization's `processInputData` should be done in ScatterPlot.tsx and LinePlot.tsx - but I'll need to take a proper look.

dmfalke:
Overall, I'm happy with this change. Both `ScatterPlot` and `LinePlot` contain relatively small amounts of code, and it makes sense to me for these to be separate components.

I also agree with @bobular's comment about the types used for the data. Generally, we should favor our data abstractions wherever possible and leave converting to a third-party type to "just-in-time".

jtlong3rd (approved):

bobular:
I am SUCH an IDIOT!!!

PR #293, opened 2022-01-25
Map viz thumbnail work

chowington:
Works toward https://github.com/VEuPathDB/web-eda/issues/784

Should be used with https://github.com/VEuPathDB/web-eda/pull/812

What's our current protocol for semantic versioning? I bumped this up a minor version because in my mind it's a new feature, but I don't know how we're currently doing things or how they generally work in 0.x semver land. UPDATE: changed to only bump the patch version in accordance with the semver discussion at the dev meeting.

### DIFF
--- a/src/map/MapVEuMap.tsx
+++ b/src/map/MapVEuMap.tsx
@@ -138,20 +138,11 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
     [screenshotter]
   );
 
-  let finalMarkers: typeof markers;
-
-  // BM: Why doesn't this work when wrapped in useEffect()?
-  // useEffect(() => {
-  // Show popups if we're in magnification mouse mode and we're not dragging.
-  // Dragging messes with our popup implementation.
-  if (mouseMode === 'magnification' && !isDragging) {
-    finalMarkers = markers.map((marker) =>
-      cloneElement(marker, { showPopup: true })
-    );
-  } else {
-    finalMarkers = markers;
-  }
-  //  }, [ markers, isDragging, mouseMode ]);
+  const finalMarkers = useMemo(() => {
+    if (mouseMode === 'magnification' && !isDragging)
+      return markers.map((marker) => cloneElement(marker, { showPopup: true }));
+    return markers;
+  }, [markers, isDragging, mouseMode]);
 
   return (
     <Map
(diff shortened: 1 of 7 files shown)


src/map/MapVEuMap.tsx
```diff
@@ -82,14 +93,63 @@ export default function MapVEuMap({
...
+        hidden: true,
+        hideElementsWithSelectors: [],
+      }),
+    []
+  );
+
+  useEffect(() => {
+    if (mapRef.current?.leafletElement)
+      screenshotter.addTo(mapRef.current.leafletElement);
+  }, [screenshotter, mapRef.current?.leafletElement]);
```
jtlong3rd, line 109: This effect is well-intentioned ("whenever the leaflet element changes, install the screenshotter plugin") but has a subtle bug.

In general, you shouldn't add a ref's `.current` value to a hook dependency array, because React doesn't track the contents of a ref. (Great explanation [here](https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/).)

There are two possible workarounds:

1. Either remove the `mapRef.current?.leafletElement` dependency, or change it to `mapRef` (as the article above explains, both approaches are valid).
2. If that doesn't work, you can move the effect logic into a [callback ref](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) on the `Map` component. Let me know if you're considering this approach - I'd be happy to go into more detail!

jtlong3rd (commented):
This is a neat approach! Just one inline suggestion came to mind.

Also, note to our future selves - there's a new major version of react-leaflet. Might be worth assessing when to upgrade. (The upgrade might be nontrivial - among other things, react-leaflet v3 appears to have dropped support for `ref`s and `useLeaflet` in favor of a new suite of `useMap*` hooks.)

src/map/MapVEuMap.tsx
chowington: Thanks for the heads-up, Jamie! That all makes sense :) Looks like just using mapRef is working 👍

jtlong3rd (approved):

bobular (approved):
All looks great.  As discussed with Connor already, I saw some embarrassing old code/comments and fixed it too.  Also some stories were still broken after I thought I'd fixed it in my previous merged PR so I added those fixes here.

Tested in EDA and looks really great!  I will add a ticket for EDA about persisting the leaflet background tile choice in user-saved analyses - and this will also need to trigger a new screenshot. Pretty sure the map component's gonna need a new prop!

Issue #295, opened 2022-01-27
Map component needs a prop for the background tile choice


bobular:
The background tile selection is currently hardcoded and I guess it can stay this way.

Applications that use this component may want to set (and remember) the background tile choice so we need a mechanism to control this via a prop, not just the leaflet UI element that we currently have.

I guess to keep things simple, the prop can be an arbitrary string.  If the value doesn't match one of the layers, then just revert to default.  There should, of course, also be a callback to communicate any state changes to the outside world.  

There's some EDA-side work here too (including updating the thumbnail generation dependencies).  Give me @bobular a shout when you get started here and I'll make a ticket there too.

chowington:
I can start on this @bobular. Do you envision keeping the leaflet UI picker as well? Would the prop just be for the default tiles or would it fully control the tiles?

chowington:
To QA, first change the map's base layer via the menu at the top right, then minimize the map viz and come back to it. The base layer should still be what you changed it to rather than the default.

PR #298, opened 2022-02-01
Map base layer prop

chowington:
Works toward https://github.com/VEuPathDB/web-components/issues/295

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.11.13",
+  "version": "0.11.14",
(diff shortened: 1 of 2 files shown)


bobular (approved):
I saw you caught some edge cases with the thumbnail capture too.

Base layer code looks and works great!

bobular:
Oh, no need to make a story for this, but please check it hasn't broken storybook's existing stories.

chowington:
> Oh, no need to make a story for this, but please check it hasn't broken storybook's existing stories.

Good thought! Seems to work fine 👍

PR #300, opened 2022-02-09
changing legend style

moontrip:
Danielle asked to change the style of custom legend for the case that [an overlay item do not have any data](https://github.com/VEuPathDB/web-eda/issues/817). Thought of too complex solution initially but found that it could be done fairly easily. 

I have confirmed with Danielle that my solution was satisfied with her: [see this comment with screenshot and her reply in the next.](https://github.com/VEuPathDB/web-eda/issues/817#issuecomment-1033185600). Thus, I just made @jtlong3rd as a sole reviewer.


moontrip:
Thanks @jtlong3rd for your quick review and approval 👍

Issue #297, opened 2022-01-28
Map component needs a `showSpinner` and possible `showNoDataOverlay` props


bobular:
Look and behaviour should match the plotly components.

nkittur-uga:
Spinner is visible. QA is complete.

PR #301, opened 2022-02-15
Add spinner and no data overlay props to map

chowington:
Works toward https://github.com/VEuPathDB/web-components/issues/297


bobular (approved):
Nice - thanks for the stories too.

PR #304, opened 2022-02-17
fix map layer issues

moontrip:
This PR will possibly resolve the issues [#840](https://github.com/VEuPathDB/web-eda/issues/840)

I think this would work, however, since UMSP data is not available, I could not test. Thus, I made this as a draft PR for the time being.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.11.15",
+  "version": "0.11.16",
   "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


moontrip:
Through Bob's new ticket, I realized that map can also be tested with SCORE Seasonal Transmission S. haematobium Cluster Randomized Trial study so I have tested my implementation. It seems to me that it works fine, so I am changing this to normal PR. Can you please review this @bobular ? Thanks in advance.

bobular (approved):
All layers working, thanks! :+1:

PR #309, opened 2022-02-24
Update from core-components -> coreui

eikonomega:


jtlong3rd (approved):

dmfalke:
Sadly, I am going to close this pull request. We are not currently using `@veupathdb/coreui`. That will likely change in the future, at which point we can use the latest version.

PR #310, opened 2022-02-28
Add missing map component dependencies

bobular:
Works towards https://github.com/VEuPathDB/web-eda/issues/875

The second one solves the inconsistent behaviour noted in 875.

I have tested the map stories and all seems to be working fine.

### DIFF
--- a/src/map/SemanticMarkers.tsx
+++ b/src/map/SemanticMarkers.tsx
@@ -134,7 +134,7 @@ export default function SemanticMarkers({
 
     // Update previous markers with the original markers array
     setPrevMarkers(markers);
-  }, [markers]);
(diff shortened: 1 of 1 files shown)


chowington (approved):
Looks good and works well!

chowington:
@bobular It's worth noting that I couldn't reproduce the issue described in the ticket on web-components main

bobular:
> @bobular It's worth noting that I couldn't actually reproduce the issue described in the ticket on web-components main

Thanks Connor. I think the double requests was fixed in web-eda, and the inconsistency was solved here, but I don't remember exactly!

bobular:
Damn I forgot to `yarn version`

PR #312, opened 2022-03-03
Chart parts

d-callan:
just something im fiddling w, mostly a proof of concept, that i thought id try to get some peoples thoughts on. at some point well have to start thinking about implementing things like violin plots, and plotly isnt likely to serve us well there. so i thought id try out a few alternatives early, starting w Microsoft chart parts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,8 @@
     "url": "https://github.com/veupathdb/web-components.git"
   },
   "dependencies": {
+    "@chart-parts/react": "^0.1.3",
+    "@chart-parts/react-svg-renderer": "^0.1.3",
     "@emotion/react": "^11.7.0",
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
--- a/src/plots/Barplot-chart-parts.tsx
+++ b/src/plots/Barplot-chart-parts.tsx
@@ -40,6 +40,12 @@ export const BarChart: FC<BarplotProps> = memo(function BarChart({
   dependentAxisRange,
   ...restProps
 }: BarplotProps) {
+  const dataSet = data.series[0];
+  const dataSetArray = dataSet.label.map((label, index) => ({
+    label: dataSet.label[index],
+    value: dataSet.value[index],
+  }));
+  const inputData = { dataSetArray };
   const [hoverIndex, setHoverIndex] = useState<number | undefined>();
   const onEnterRect = useCallback(
     ({ index }) => {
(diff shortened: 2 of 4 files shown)


moontrip:
Interesting to see MS chart in react 👍 Certainly there will be limitations on plotly.js and I don't know we'll what plots we will need further, but just FYI, there is a violin plot in plotly.js, https://plotly.com/javascript/violin/. Though its functionality may not meet your needs 😃

d-callan:
Yea. My understanding of the violin trace in plotly is that it can't accept precomputed values. It seems to insist on being provided raw data and determining kernel density estimates itself. I considered for a minute maybe making a custom trace or something. But then I thought it's unlikely that violin will be our only case of this type of problem, because the problem really is that plotly exists at a level of abstraction that has turned out in hindsight not to be flexible enough for us. Chart parts is more abstracted than plotly and intended for use w react, so it seemed a strong enough option to at least try. Here's the github project if anyone is interested: https://github.com/microsoft/chart-parts

d-callan:
I also wonder if anyone has thoughts on what else to try? I figure visx is the other obvious one. That is more abstracted yet than chart parts, which may have some pros and cons. But others?

d-callan:
One last thought before I forget.. I think accessibility may play in to this decision. It may be that once we know what we need in terms of accessibility for our plot components that the appropriate level of abstraction will become obvious.

moontrip:
Thank you for sharing your thought and experience, @d-callan 👍 Although I am not an advocate of plotly.js, for violin plot, they may accept precomputed values: for sure I did not test it though 😞 The reason why I say this comes from my past experience with boxplot, which I thought they only accept raw data but realized that they also work with precomputed values (e.g., q1, q3, upper/lowerfence, etc.) - still hard to find that aspect from their site.

But I agree that plotly.js may not be that flexible than other packages. Honestly I did not know well about MS chart parts but I am a bit concerned purely based on its current version (v0.1.3): so I may be wrong. I also remember that Mike applauded visx (😄) and we are using it for the entity diagram

asizemore:
@d-callan what all does "accessibility in plot components" encompass? My general thought is text size, contrast, good colormap, etc. Lots of those i think we're hitting well. Are there other aspects i'm not thinking of?

d-callan:
Things like alt text and tab able elements come to mind... but it's a bit of an open question yet what exactly being accessible would mean for us.

bobular:
Just stopped by to say I've looked at the code and it looks nice. Cleaner than Plotly

Does it support openGL for large scatter plots?

PR #315, opened 2022-03-08
add scale to map and change map mouse tool for mapViz

### LINKED ISSUE #313: Map: MapVEuMap should show a scale bar
Like you see bottom right in the legacy map.

image

Location should probably also be bottom right.  Add a prop to turn on/off.

moontrip:
This resolves #313 

add a prop, showScale, to control for displaying (distance) Scale in the map. Default is set to be true for now.

For test purpose, I have changed the Default story at AnimatedMarkers story to check the behavior of showScale: a) add setViewport at useState(); b) add onViewportChanged function to track the change of zoom level change; c) add showScale prop at Default story with condition. It worked fine as expected (it would be shown when zoom > 4).

For a note to use at MapViz (at web-eda), I have tested the showScale condition. It should be something like this:

when calling `<MapVEuMap>`
`showScale={zoomLevel != null && zoomLevel > 4 ? true : false}`  

update: I have made a [PR at web-eda](https://github.com/VEuPathDB/web-eda/pull/949) to implement above suggestion. Confirmed that it worked fine - you can test by zooming out and in.


moontrip:
@bobular map mouse tool is needed to be change so that it can be used with mapViz at web-eda. 
I have added this work in this branch so that you can test them (scale and mouse tool) at the same time.

It appears to me that fontawesome at web-eda is old-fashioned v4 while web-components is v5. Thus, map mouse tool's icons were made with v5 which was not correctly shown in mapViz. Another issue I found when I was working for react-leaflet v3 upgrade is that mouseover rectangle bounds sometimes conflicts with mouseover popup. Thus I disabled mouseover rectangle bounds when popup is supposed to be shown (magnifying icon at mouse tool). This can be tested at the same [web-eda PR](https://github.com/VEuPathDB/web-eda/pull/949) that now contains scale and mouse tool.

Here is a screenshot
mouse tool1

bobular (approved):
I approve the changes here, but as I mentioned [here](https://github.com/VEuPathDB/web-eda/pull/949#issuecomment-1064320798) there is a problem with the calculation of "am I near the edge of the map" when deciding which direction the enlarged markers should pop "up".

That could be a new issue if not easily solved here.  It would be nice to have in the demo!

PR #317, opened 2022-03-09
Fix styling of faceted plots with no data

jernestmyers:
Addresses issue [#925 in web-eda](https://github.com/VEuPathDB/web-eda/issues/925).

- Passed the optional opacity props to `NoDataOverlay` and set its value to 0.85.
- Changed the color of the "No data" message to be one shade darker.

image


bobular (approved):
Looks good to me! Thanks.

Your screenshot illustrates the issue with long variable values (e.g. "Enrollment, outcome and additional blah blah"). They get wrapped over to the next line(s) and I think that's fine for now, even though it doesn't look super polished.

jernestmyers:
Closes VEuPathDB/web-eda#925

PR #318, opened 2022-03-09
Fix broken selection in Histogram::Range Selection story

bobular:
The `valueType` was missing, so the blue rectangle shape was half number, half date, and Plotly was quietly refusing to plot it.


jernestmyers (approved):
Ah, I recall running into a Histogram that wouldn't do the blue selector when I was working on my ticket. As I'm so new, I wasn't certain that it wasn't intentional. I'll see if I can find it and check if it's the same issue!

bobular:
> Ah, I recall running into a Histogram that wouldn't do the blue selector when I was working on my ticket. As I'm so new, I wasn't certain that it wasn't intentional. I'll see if I can find it and check if it's the same issue!

Cool. This is in Storybook (just in case that wasn't clear)

bobular:
Just going to merge this without a version update because it doesn't affect any client code.

Issue #321, opened 2022-03-10
Map: mouse-over magnified markers don't always assume correct orientation (when near map edge)


bobular:
When deployed as an EDA visualization, the MapVEuMap's BoundsDriftMarkers are not "popping-left/down/right" when near the edges of the map.  If you resize your browser window (smaller) and move the map to match the browser window, it seems to work.  It's something to do with div sizes and percentages, and @moontrip has diagnosed and [written more here](https://github.com/VEuPathDB/web-eda/pull/949#issuecomment-1064352960).

Ideally, we would create a story with a similar map embedding as in EDA and reproduce the error.  Then fix it (in BoundsDriftMarker) :-)

PR #323, opened 2022-03-11
Fix popup reorientation

chowington:
Works toward https://github.com/VEuPathDB/web-components/issues/321

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.11.26",
+  "version": "0.11.27",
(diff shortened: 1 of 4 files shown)


moontrip:
@chowington I have a chance to test this via web-eda's mapViz (using yarn link) and it works well 👍 (Didn't test story but based on mapViz test results it would be just fine I suppose 😄).

moontrip (approved):
This works for me for mapViz. Can you please check this PR @bobular as well ?

bobular (approved):
Tested in EDA and the Storybook, and the code looks great. Thanks!

PR #328, opened 2022-03-22
move "dependencies" to "{peer,dev}Dependencies"

dmfalke:
refs VEuPathDB/web-dev#11


jtlong3rd (approved):
Makes sense! I'm guessing that in general, good candidates for `peerDependencies` will include (but not be limited to):

1. Our in-house `@veupathdb` packages
2. Packages introducing context and other objects which need to be unique
3. React and common `react-*` libraries

dmfalke:
> Makes sense! I'm guessing that in general, good candidates for `peerDependencies` will include (but not be limited to):
> 
> 
> 
> 1. Our in-house `@veupathdb` packages
> 
> 2. Packages introducing context and other objects which need to be unique
> 
> 3. React and common `react-*` libraries

I think that's a good rubric!

PR #331, opened 2022-03-24
More flexible tooltip text and properly allow nulls in scatter/line data

bobular:
Just one review (of the three requested) would be nice thanks!


moontrip (approved):
Thank you for doing this 👍

Issue #322, opened 2022-03-11
Histogram: right hand coordinate of blue selection area seems to be affected by local time zone for date variables


bobular:
In EDA subsetting, if you zoom in to few days of GEMS1 observation date, binning at 1 day, and make a selection, the right hand edge of the blue selection is not quite at the edge of the bin.

![image](https://user-images.githubusercontent.com/308639/157882139-28042090-72ce-483b-934c-bb03361c45c9.png)

It turns out that the right hand edge isn't drawn at 1 millisecond to midnight, but 1 millisecond to 11pm

`x1: "2011-05-02T22:59:59.999Z"`

I am in GMT/UTC timezone, which makes this super-mysterious.

For those on Eastern time, the `x1` coordinate is even more displaced.

Here's the code that is responsible.

https://github.com/VEuPathDB/web-components/blob/9c69105b4948585019738de644ccfd47b7a194c6/src/plots/Histogram.tsx#L297-L300

It may be that `new Date(range.max)` is made in the local time. `range.max` may need 'T00:00:00Z' appending before converting to Date.

While diagnosing/testing it may be necessary to "spoof" different timezones from your browser. Chrome dev tools can do this, as can firefox plugins (not specifically dev-oriented).

SheenaTomko:
<img width="826" alt="Screen Shot 2022-04-11 at 3 10 32 PM" src="https://user-images.githubusercontent.com/54597319/162823992-82fc3b0b-7b10-4d90-a9e6-5d0f88b805f9.png">

This looks good now. Also tested with San Francisco, Tokyo, and a couple other timezones. No issues

PR #332, opened 2022-03-25
Improve handling of date variables in blue selection range of histograms

jernestmyers:
Addresses issue #322.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.14.0",
+  "version": "0.14.1",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
--- a/src/plots/Histogram.tsx
+++ b/src/plots/Histogram.tsx
@@ -231,42 +231,6 @@ const Histogram = makePlotlyPlotComponent(
     // local state for range **while selecting** graphically
     const [selectingRange, setSelectingRange] = useState<NumberOrDateRange>();
 
-    /* !---- OVERRIDES THE LOGIC THAT PLOTLY USES FOR LAST BIN WITH DATES ----!
-    This block calculates a date that is the midpoint of the lastBin's binMiddle and binEnd
-    This midpoint is used as the threshold that determines if the selection should snap
-    to lastBin.binEnd
-      **  It's still unclear to me why we subtract 1 day. After further thought, I think most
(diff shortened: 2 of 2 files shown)


jernestmyers:
@bobular I'm curious what you think about this approach. I used Chrome's dev tools to play with time zones and didn't notice any issues thus far.

It seems like `range.max` already has the date format necessary for `rightCoordinate` since my previous change uses the last bin's `binEnd` as the limit irrespective of format.

bobular (commented):
Hi! When I am "zoomed" in to day resolution (GEMS1 "observation date"), the right-most bin is "snapping" properly, but the others only select/go blue if you drag past the half-way point of the bin/bar to the right of the one you are selecting. So to select one day, you have to drag over two bars.

In week-resolution mode, the snapping is behaving fine, but the blue rectangle isn't wide enough.

Looks like some more testing/discussion may be required - we could meet for this?

(Just checking, there is no `web-eda` PR that goes with this?)

jernestmyers:
@bobular I apologize for not pushing my most up-to-date changes until this morning. As a result, I'm not sure if you tested the most recent code and read those comments. Either way, that behavior you described remains and it is due to how we subtract 1 day from the selector. I am not sure why it was ever necessary, but I feel that removing that adjustment will dramatically simplify my approach.

bobular:
I think the problem is that `DateMath.endOf()` is calculating the end of the day in the browser's local time zone (for the date in question, so the May 2011 observation dates are in BST (GMT+1) for me)

e.g. Sun May 08 2011 23:59:59 GMT+0100 (British Summer Time)

When converted back to ISO string it becomes 2011-05-08T22:59:59.999Z

I don't see an obvious solution, other than to add on the offset, which you can get from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset

bobular:
Or (if non-integers allowed, which I assume they are) add 0.9999999 days.

jernestmyers:
While testing with multiple timezones, I noticed that those with `timezoneOffsets > 0` were behaving as I would have expected. But those with negative `timezoneOffsets` were jumping ahead a day during selection. I am not entirely sure as to why that happens, but subtracting an extra day's worth of minutes seemed to solve the problem.

bobular (approved):
Works well for me in BST-land!

Issue #326, opened 2022-03-03
Update entity diagram to differentiate 1:1 vs 1:many relationships


danicahelb:
There has been a proposal to update the entity diagram. Arrows should be removed as there is no directionality in relationships. Nodes can be styled to differentiate 1:1 vs 1:many relationships.

Mouse-over text for the edges? May be difficult to universalize accurately…
For example:
- In this study, each Household had one Participant
- In this study, each Household had one or more Household Repeated Measures

<img width="666" alt="image" src="https://user-images.githubusercontent.com/25450900/156658713-947bd475-70f0-4609-9a4a-d9fb3956174f.png">

This proposal needs sign-off from EDA WG before implementing...

SheenaTomko:
This file contains all 1:1 relationships that need to be updated in the database. 

[1_to_1_relationships.csv](https://github.com/VEuPathDB/web-components/files/8496532/1_to_1_relationships.csv)

SheenaTomko:
The lines themselves look good. Ryan changed all to be 1:1 for this release since dd isn't ready for the change. Will have a separate redmine ticket for DD to implement the schema changes and load this 1:many information as part of the workflow

danicahelb:
For b57, we have successfully replaced the arrows with nodes
<img width="1099" alt="image" src="https://user-images.githubusercontent.com/25450900/163886834-20ebef46-0a8c-4bed-9afc-5966b9423134.png">

We are not yet able to differentiate 1:1 vs 1:many relationships with the diagram. Backend work is needed, see: https://redmine.apidb.org/issues/47686

PR #336, opened 2022-04-01
Update entity diagram UI to represent 1:1 and 1:many relationships

jernestmyers:
Addresses issue #326

Current code is hardcoded to only render repeated measures as a 1:many relationship. I will update when backend includes the necessary fields/properties.

The current implementation looks as follows:
**Amazonia ICEMR Brazil Cohor**
image

**GEMS1 Case Control**
image

### DIFF
--- a/src/EntityDiagram/EntityDiagram.stories.tsx
+++ b/src/EntityDiagram/EntityDiagram.stories.tsx
@@ -18,10 +18,11 @@ const rootEntity: StudyData = {
   description: 'Households from the study area',
   children: [
     {
-      id: 'EUPATH_0000776',
+      id: 'GEMS_HouseObs',
       displayName: 'Household Observation',
       description: '',
       children: [],
+      isManyToOneWithParent: true,
       variables: [
         {
           id: 'var-19',
--- a/src/EntityDiagram/EntityDiagram.stories.tsx
+++ b/src/EntityDiagram/EntityDiagram.stories.tsx
@@ -18,7 +18,7 @@ const rootEntity: StudyData = {
   description: 'Households from the study area',
   children: [
     {
-      id: 'GEMS_HouseObs',
+      id: 'EUPATH_0000776',
       displayName: 'Household Observation',
       description: '',
       children: [],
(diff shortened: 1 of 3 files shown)


asizemore:
Hey @jernestmyers !! Just checking out the screenshots it looks really good!

Can we please decrease the stroke width? Maybe like strokeWidth of 2 and nodeSize when expanded 4.5? (my hack to keep the circle the same size but decrease the stroke width)

jernestmyers:
Thanks for the feedback! I'll make those adjustments this morning.

jernestmyers:
@asizemore I ended up reducing the `nodeSize` to 4.25 after integrating the new 1:many boolean data. Every entity I have checked thus far has `isManyToOneWithParent` set to `true` and `nodeSize = 4.5` felt a bit big in that case. But I'm happy to bump it back up to 4.5 if that's preferred.

**Amazonia Brazil**
image

**GEMS 1 Case Control**
image

src/EntityDiagram/EntityDiagram.stories.tsx
```diff
@@ -164,7 +165,7 @@ const entityCounts: EntityCounts = {
     total: 100,
     filtered: 5,
   },
-  GEMS_HouseObs: {
+  EUPATH_0000776: {
```
dmfalke, line 168: The property name on this line should also be changed to `EUPATH_0000776`:
https://github.com/VEuPathDB/web-components/blob/346a968e647ca37ef539dca819e697c25076d050/src/EntityDiagram/EntityDiagram.stories.tsx#L21. As it is, I get an error in the storybook when switching the size.

dmfalke (changes requested):
All of this looks great, with the exception of one issue, for which I left a comment.

src/EntityDiagram/EntityDiagram.stories.tsx
jernestmyers: Ah, I had changed from `GEMS_HouseObs` to `EUPATH_0000776` because the code was using the `id` property to determine if something was 1:many. Now that we rely on `isManyToOneWithParent`, I was changing it back to `GEMS_HouseObs` but forgot to replace it there as well.

Would you agree I should just go back to `GEMS_HouseObs` in that case for consistency?

dmfalke: Yeah, that sounds good!

asizemore:
@jernestmyers the 4.25 size looks great! Good call!

dmfalke (approved):

PR #337, opened 2022-04-07
changing control text color to black

moontrip:
This PR related to https://github.com/VEuPathDB/web-eda/issues/1011

Danica asked me to change control's text colors into black, not light gray.

### DIFF
--- a/src/components/widgets/NumberAndDateInputs.tsx
+++ b/src/components/widgets/NumberAndDateInputs.tsx
@@ -182,7 +182,8 @@ function BaseInput({
       {label && (
         <Typography
           variant="button"
-          style={{ color: focused ? DARK_GRAY : MEDIUM_GRAY }}
+          // style={{ color: focused ? DARK_GRAY : MEDIUM_GRAY }}
+          style={{ color: 'black' }}
         >
           {label}
         </Typography>
--- a/src/components/widgets/NumberAndDateInputs.tsx
+++ b/src/components/widgets/NumberAndDateInputs.tsx
@@ -182,8 +182,7 @@ function BaseInput({
       {label && (
         <Typography
           variant="button"
-          // style={{ color: focused ? DARK_GRAY : MEDIUM_GRAY }}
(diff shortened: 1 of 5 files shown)


src/components/widgets/NumberAndDateInputs.tsx
```diff
@@ -182,7 +182,8 @@ function BaseInput({
       {label && (
         <Typography
           variant="button"
-          style={{ color: focused ? DARK_GRAY : MEDIUM_GRAY }}
+          // style={{ color: focused ? DARK_GRAY : MEDIUM_GRAY }}
+          style={{ color: 'black' }}
```
bobular, line 186: Please could you make all these like the following:

```
  color: disabled ? MEDIUM_GRAY : 'black'
```
See BinWidthControl.tsx and Switch.tsx for prior examples.

Thanks!

bobular (changes requested):

src/components/widgets/NumberAndDateInputs.tsx
moontrip: @bobular Thank you for your suggestion! 👍 I missed that point. Note that I also changed NumberAndDateRangeInputs component with the same approach for "to" text haha. Now texts are shown in medium gray when X axis variable is ordinal/categorical - screenshot is attached for example.

lineplot-ordinal

moontrip:
@bobular Thank you for taking care of the changes in your `Slider` component/widget! 👍 I saw that max() part at the widget but I was hesitant to mess up your calculation in there without in-depth knowledge. I made changes following your suggestions at web-components, https://github.com/VEuPathDB/web-components/pull/337/commits/7a916be60f6abaa3e8d584fbf8fc4a7e41573c3a, and it seems to work great! Here is my screenshot on a lineplot case and now bin width is in the single line!

binWidth

bobular (approved):
Looks great. No need for the `// add nowrap` or `// simplyfying width` comments, in my opinion.

PR #338, opened 2022-04-11
Switch chart markers to use barplot; add display labels to RadioButtonGroup

bobular:
The enlarged popups for ChartMarkers now use Barplot instead of Histogram. 
Also allowed an array of colors (per series) to be passed to Barplot. This should not affect regular barplots.
Also added optionLabels to RadioButtonGroup - I added a Story for this but was unable to get Storybook running on my laptop (white browser window - no errors in the terminal (forgot to check the FF console)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.14.5",
+  "version": "0.15.0",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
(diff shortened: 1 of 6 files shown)


moontrip:
Looks good except that Chart Marker stories here do not work: Map viz seems to work fine though.

bobular:
> Looks good except that Chart Marker stories here do not work: Map viz seems to work fine though.

I've fixed this now. Please could you re-review?

moontrip (approved):
@bobular yes the story seems to work fine now 👍

PR #340, opened 2022-04-12
Introduce TypeScript-based map config (resolves #324)

### LINKED ISSUE #324: Fix build errors when bundling with webpack v5
With webpack version 5, named imports from json modules are not supported. This is a breaking change from version 4. This change is needed for https://github.com/VEuPathDB/web-dev/pull/10.

There are two options for addressing this:
1. Replace named imports with default imports.
2. Replace json modules with TypeScript modules.

### Replace named imports with default imports

An example of a named import from a json module is
```
import { defaultAnimationDuration } from './map.json';
```

This can be converted to a default import:
```
import mapConfig from './map.json';
const { defaultAnimationDuration } = mapConfig;
```

### Replace json modules with TypeScript modules

`map.json` may look something like
```json
{
    "defaultAnimationDuration": 400
}
```

This can be converted to a TypeScript module with a named export
```typescript
export const defaultAnimationDuration = 400;
```

The named export can then be consumed like
```typescript
import { defaultAnimationDuration } from './map';
```

jtlong3rd:
Resolves #324.

I have left the old `map.json` in place, so as to avoid making a breaking change. We should remove `map.json` in the next major release of `@veupathdb/web-components`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.14.4",
+  "version": "0.14.5",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
--- a/src/map/config/map.ts
+++ b/src/map/config/map.ts
@@ -1,0 +1,35 @@
+// N.B.: This config should be used in place of ./map.json,
+// which is deprecated.
+// (Context: https://github.com/VEuPathDB/web-components/issues/324)
+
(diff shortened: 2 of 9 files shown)


dmfalke (approved):
*sigh* I suppose removing it would incur a major version update.

jtlong3rd:
> _sigh_ I suppose removing it would incur a major version update.

I'll open an issue for removing `map.json` during the next major version update. (Whenever that ends up being.)

PR #314, opened 2022-03-06
upgrading to react-leaflet v3 and corresponding changes etc

moontrip:
This is an attempted to adopt react-leaflet v3.

- TL:DR
From my tests, this works for all map related stories and functionality implemented at the map. Revising codes to work with this upgrades. Tidying some codes, especially Donut/Chart makers.

- To-do
a) leaflet upgrade is not applied yet. Will do that.
b) Some functionality regarding ref, e.g., flyTo and thumbnail could not be tested as it should be done through web-eda's mapViz: I have tried to make it happen but untested. Thus, it would be done with web-eda.

- Installation note
There are many changes in packages so it is recommended to remove ./node_modules directory and yarn install for this work, just in case.

- some details for compiling issue
a) add/upgrade packages:
`react-leaflet` v2 -> v3
`@types/leaflet`
`react-leaflet-drift-marker`: replacing `leaflet-drift-marker`
`@babel/plugin-proposal-nullish-coalescing-operator`

b) changes for successful compiling, including storybook
i) nullish coalescing operator issue from react-leaflet v3 and CRA
browserslist at package.json:
from
`
  "browserslist": [
    "extends @veupathdb/browserslist-config"
  ],
`
to
`
  "browserslist": [
    ">0.2%",
    "not dead",
    "not op_mini all"
  ],

`
ii) changed `.storybook/main.js` for running storybook correctly

PR #347, opened 2022-04-28
fixing ref bug for flyTo and thumbnail

moontrip:
This branched out from react-leaflet-v3-test to fix ref bug for flyTo and thumbnail. Also made a story to test them at web-components: 

In the story, flyTo delay is set to be a bit long (2000 ms) to see its visual effect. A screenshot to show thumbnail is as follows. Note that map and thumbnail sizes are not set to be actual ones as it is a demo purpose and large thumbnail takes too much time to decode. 

map-thumbnail-story-test

### DIFF
--- a/src/map/MapVEuMap.tsx
+++ b/src/map/MapVEuMap.tsx
@@ -23,6 +23,8 @@ import {
   LayersControl,
   ControlledLayerProps,
   ScaleControl,
+  // import useMap hook
+  useMap,
 } from 'react-leaflet';
 import { SimpleMapScreenshoter } from 'leaflet-simple-map-screenshoter';
 import SemanticMarkers from './SemanticMarkers';
@@ -205,9 +207,10 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
   // Whether the user is currently dragging the map
   const [isDragging, setIsDragging] = useState<boolean>(false);
 
-  // This may be useRef<MapContainer> but need to check with below useEffect comment
-  const mapRef = useRef<any>(null);
+  // set useSatate to handle map instance instead of useRef using 'whenCreated' prop at MapContainer
+  const [mapRef, setMapRef] = useState<any>(null);
 
+  // screenshot part
   const screenshotter = useMemo(
     () =>
       new SimpleMapScreenshoter({
@@ -217,10 +220,10 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
     []
   );
 
-  // this needs to be checked with map Viz as it appears .leafletElement does not exist in v3
   useEffect(() => {
-    if (mapRef.current?.leafletElement)
-      screenshotter.addTo(mapRef.current.leafletElement);
+    if (mapRef) {
+      screenshotter.addTo(mapRef);
+    }
   }, [screenshotter, mapRef]);
 
   useImperativeHandle<PlotRef, PlotRef>(
@@ -232,8 +235,7 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
           // Wait to allow map to finish rendering
           await new Promise((resolve) => setTimeout(resolve, 1000));
 
-          // Check that map leaflet element still exists
-          if (mapRef.current) {
+          if (mapRef) {
             // Call the 3rd party function that actually creates the image
             const screenshot = await screenshotter.takeScreen('image', {
               domtoimageOptions: {
@@ -241,6 +243,7 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
                 height: imageOpts.height,
               },
             });
+
             // The screenshotter library's types are wrong. TS thinks this next line
             // will never happen, but takeScreen('image') should in fact return a string
             if (typeof screenshot === 'string') return screenshot;
@@ -254,8 +257,99 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
         return '';
       },
     }),
-    [screenshotter]
+    [screenshotter, mapRef]
   );
+
+  const finalMarkers = useMemo(() => {
+    if (mouseMode === 'magnification' && !isDragging)
+      return markers.map((marker) => cloneElement(marker, { showPopup: true }));
+    return markers;
+  }, [markers, isDragging, mouseMode]);
+
+  return (
+    // change from Map to MapContainer
+    <MapContainer
+      center={viewport.center}
+      zoom={viewport.zoom}
+      style={{ height, width, ...style }}
+      className={mouseMode === 'magnification' ? 'cursor-zoom-in' : ''}
+      minZoom={1}
+      worldCopyJump={false}
+      ondragstart={() => setIsDragging(true)}
+      ondragend={() => setIsDragging(false)}
+      onbaselayerchange={(event: ControlledLayerProps) =>
+        onBaseLayerChanged && onBaseLayerChanged(event.name as BaseLayerChoice)
+      }
+      // this prop is used to use map instance
+      whenCreated={setMapRef}
+    >
+      {/* PerformFlyToMarkers component for flyTo functionality */}
+      {flyToMarkers && (
+        <PerformFlyToMarkers
+          markers={markers}
+          flyToMarkers={flyToMarkers}
+          flyToMarkersDelay={flyToMarkersDelay}
+        />
+      )}
+
+      <TileLayer
+        url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
+        attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
+      />
+
+      <SemanticMarkers
+        onBoundsChanged={onBoundsChanged}
+        markers={finalMarkers}
+        animation={animation}
+        recenterMarkers={recenterMarkers}
+      />
+
+      {showMouseToolbar && (
+        <MouseTools mouseMode={mouseMode} setMouseMode={setMouseMode} />
+      )}
+
+      {showGrid && zoomLevelToGeohashLevel ? (
+        <CustomGridLayer zoomLevelToGeohashLevel={zoomLevelToGeohashLevel} />
+      ) : null}
+
+      <LayersControl position="topright">
+        {Object.entries(baseLayers).map(([name, layerProps], i) => (
+          <LayersControl.BaseLayer
+            name={name}
+            key={name}
+            checked={baseLayer ? name === baseLayer : i === 0}
+          >
+            <TileLayer {...layerProps} />
+          </LayersControl.BaseLayer>
+        ))}
+      </LayersControl>
+
+      {showSpinner && <Spinner />}
+      {showNoDataOverlay && <NoDataOverlay opacity={0.9} />}
+      {/* add Scale in the map */}
+      {showScale && <ScaleControl position="bottomright" />}
+    </MapContainer>
+  );
+}
+
+export default forwardRef(MapVEuMap);
+
+// for flyTo
+interface PerformFlyToMarkersProps {
+  /* markers */
+  markers: ReactElement<BoundsDriftMarkerProps>[];
+  /** Whether to zoom and pan map to center on markers */
+  flyToMarkers?: boolean;
+  /** How long (in ms) after rendering to wait before flying to markers */
+  flyToMarkersDelay?: number;
+}
+
+// component to implement flyTo functionality
+function PerformFlyToMarkers(props: PerformFlyToMarkersProps) {
+  const { markers, flyToMarkers, flyToMarkersDelay } = props;
+
+  // instead of using useRef() to the map in v2, useMap() should be used instead in v3
+  const map = useMap();
 
   const markersBounds: MapVEuBounds | null = useMemo(() => {
     if (markers) {
@@ -302,9 +396,9 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
         [ne.lat + latBuffer, ne.lng + lngBuffer],
       ]);
 
-      mapRef.current?.leafletElement?.fitBounds(boundingBox);
+      map.fitBounds(boundingBox);
     }
-  }, [markersBounds, mapRef]);
+  }, [markersBounds, map]);
 
   useEffect(() => {
     const asyncEffect = async () => {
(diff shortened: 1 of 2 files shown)


moontrip:
Just for an information.

TL;DR (@dmfalke? 😄)

In react-leaflet v2 we have used useRef() hook to access map for flyTo, thumbnail, etc. However, in v3, ref is not supported so a new method is required. Basically there are two ways to replace useRef().

a) use MapContainer's whenCreated prop, e.g.,
```
const [mapRef, setMapRef] = useState(null);
...
<MapContainer
  ...
  whenCreated={setMapRef}
  ...
/>
``` 

In this manner, mapRef becomes map instance so we can use it for accessing map. thumbnail fix in this PR utilizes this approach.  A caution is that some users commented that the use of whenCreated may not be consistent, though I did not notice it in this case.

b) use child component and a new useMap() hook introduced in v3, e.g.,
```
function MapVEu(props) {
  ...
  return (
    <MapContainer>
      <PerformFlyToMarkers>
    </MapContainer>
  ); 
}

function PerformFlyToMarkers(props) {
   const mapRef = useMap();       // now map can be accessed via mapRef
   mapRef.fitbounds(bounds)       // fitbounds()
   ...
   return null; 
}
```

It works well as I used this approach for flyTo functionality. Some notes are: a) useMap() cannot be used in the component that returns <MapContainer> (in this example, MapVEu component). Thus, a child component is needed to utilize useMap() like above example code; b) if forwardRef() is necessary, e.g., at <PerformFlyToMarkers>, it means that forwardRef should be used twice in this case, like forwardRef(MapContainer) and forwardRef(PerformFlyToMarkers).

dmfalke:
Thanks for the details, @moontrip. Either approach seems fine to me, assuming `whenCreated` has consistent behavior.

moontrip:
> Thanks for the details, @moontrip. Either approach seems fine to me, assuming `whenCreated` has consistent behavior.

@dmfalke As I mentioned earlier, it seems to fine for our application. I saw a user's comment concerning inconsistent behavior at react-leaflet when using `whenCreated`, but it might have been revised since then as v3.2.5 (the latest v3) dealt with something regarding whenCreated - don't know details though, just looked at change log.

bobular:
Unrelated to the fly-to functionality - I am seeing new compiler warnings when building/launching storybook

```
DA/web-components/node_modules/comma-separated-tokens/index.js[BABEL]
Note: The code generator has deoptimised the styling of /home/maccallr/Desktop/EDA/web-components/node_modules/mapbox-gl/dist/mapbox-gl-unminified.js as it exceeds the max of 500KB.
```
This seems to happen for many different modules

bobular (approved):
This looks good and the story works for me too.

I tried npm-pack-here in the latest EDA and got a runtime error when adding a variable to the map visualization.

```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports. Check the render method of `BoundsDriftMarker`.
```

Has the interface to BoundsDriftMarker changed?  I can't see in this particular PR. So I'll approve this PR and then look at the main draft PR. All the other visualizations seem to work fine!

bobular:
I couldn't find an obvious issue with BoundsDriftMarker - the error message suggested that in the code there was something like

```
  return <Foo ...> </Foo>;
```
where `Foo` is undefined.  Could possibly be a problem with npm-pack-here, but everything else works.  It's a mystery.  @dmfalke or @jtlong3rd would you be able to test/diagnose this please?  (EDA `main` branch with npm-pack-here pulling in  `web-components/react-leaflet-v3-ref ` then try to make a map viz in SCORE Seasonal...)

moontrip:
> This looks good and the story works for me too.
> 
> I tried npm-pack-here in the latest EDA and got a runtime error when adding a variable to the map visualization.
> 
> ```
> Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports. Check the render method of `BoundsDriftMarker`.
> ```
> 
> Has the interface to BoundsDriftMarker changed? I can't see in this particular PR. So I'll approve this PR and then look at the main draft PR. All the other visualizations seem to work fine!

@bobular Thank you for your review and tests! 👍 Certainly there was some changes on `BoundsDriftMarker` component as noted in the parent PR's comment #314, `react-leaflet-drift-marker: replacing leaflet-drift-marker`. 

Off the top of my head... When I was implementing v3, I have removed the following:

```
const FixedDriftMarker = DriftMarker as React.ComponentClass<
  ExtractProps<typeof DriftMarker> & LeafletMarkerProps
>;
```

and used `ReactLeafletDriftMarker` directly (from `react-leaflet-drift-marker` package):
i.e., instead of `<FixedDriftMarker>`, used `<ReactLeafletDriftMarker>`

Since the relevant story file (AnimatedMarkers.stories.tsx) seemed to work fine, I thought it would be alright without that. So, just in case please check if that is the case. 

p.s. Any specific configuration you did with npm-pack-here? I couldn't make it before with a compiling error in my case: kind of `can't find MapContainer` error which the `MapContainer` replaces `Map` component in v3. So it means that the compiler tried to to find that component via v2, not v3.

bobular:
Hmm, so `node_modules/react-leaflet/package.json` should not show 2.8.0 as the version, right?

I didn't do anything special with npm-pack-here, just 

```
npm-pack-here --target ../web-components
yarn add -D file:local_modules/@veupathdb/components
yarn install --check-files
```

I'm seeing 
```
web-eda/node_modules/@veupathdb/eda/node_modules/@veupathdb/components/package.json:    "react-leaflet": "^2.7.0",
```
which doesn't look good. What's `@veupathdb/eda`? I'll try removing node_modules and start again

moontrip:
@bobular thank you for your information: basically that is what I did and failed to compile :( 
I remember that Dave made a comment on that stuff showing new and older versions at the same time, concerning npm-pack-here at our dev call.

Regarding What's @veupathdb/eda? Please see Jamie's great explanation https://github.com/VEuPathDB/web-eda/pull/1117#issuecomment-1116436946

bobular:
Thanks I will read that. In the meantime...
```
yarn why v1.22.11
[1/4] Why do we have the module "react-leaflet"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "react-leaflet@3.2.5"
info Reasons this module exists
   - "@veupathdb#components" depends on it
   - Hoisted from "@veupathdb#components#react-leaflet"
info Disk size without dependencies: "6.98MB"
info Disk size with unique dependencies: "7.2MB"
info Disk size with transitive dependencies: "7.2MB"
info Number of shared dependencies: 1
=> Found "@veupathdb/eda#react-leaflet@2.8.0"
info Reasons this module exists
   - "@veupathdb#eda#@veupathdb#components" depends on it
   - Hoisted from "@veupathdb#eda#@veupathdb#components#react-leaflet"
```

bobular:
So... `@veupathdb/eda` is the package exported by the `web-eda` repo.

To build the dev react server from `web-eda`, we shouldn't need to import this module.

But... we do import `@veupathdb/web-common` as a devDependency

After some hunting, `@veupathdb/web-common` is the npm module exported from https://github.com/VEuPathDB/EbrcWebsiteCommon/tree/master/Client, which has peer and devDependencies for `@veupathdb/eda`

So I guess that's the circular dependency that @jtlong3rd mentioned in the link from DK, because `@veupathdb/eda` has a dev dependency for `@veupathdb/components`

I have a feeling it would work if I checked out web-common, and npm-pack-here'ed web-components into it, and then npm-pack-here'ed web-common into web-eda... Let's see

bobular:
Damn, that doesn't work
```
npm-pack-here -t ../../web-components
EbrcWebsiteCommon/Client$ yarn add -D file:local_modules/@veupathdb/components
error An unexpected error occurred: "expected hoisted manifest for \"@veupathdb/components#react-leaflet\"".
```
out of my depth now!

bobular:
I just saw that `web-eda` has a devDependency for `@veupathdb/eda` which is ITSELF.

The reason for this is that `web-common`'s
https://github.com/VEuPathDB/EbrcWebsiteCommon/blob/master/Client/src/WorkspaceRouter.tsx
does this:
```
export { WorkspaceRouter as default } from '@veupathdb/eda/lib/workspace/WorkspaceRouter';
```
And not sure this is my area of expertise ;-)

bobular:
> And not sure this is my area of expertise ;-)

But I soldiered on anyway. I think I found a way to get around the eda->eda circular dependency with npm-pack-here.

But then I got into some parsing issues. @moontrip, did you say the new react-leaflet was a using a different version of ES or something?

moontrip:
@bobular As far as I know, the v3 targets ES2019. I did not experience a specific issue at web-components for compiling, except one case, nullish coalescing operator issue - had to change some configurations that are listed at https://github.com/VEuPathDB/web-components/pull/314#issue-1160644278. 

Not quite sure what the parsing issues you meant. Anyway sounds great if you found the way 👍

bobular:
Yes @moontrip it was a nullish coalescing operator thing.

I think see how to fix it on the EDA side.

Now I will try to repeat (and document) how to test this PR in EDA!

dmfalke:
I'm looking into the general issues with testing this. Specifically: circular dependency and newer syntax.

dmfalke:
I was able to get things working by doing to following:

moontrip:
I tried to test Dave's change on the package.json (web-eda) with the npm-pack-here yesterday. It worked fine without specific error I noticed. By doing this so, only react-leaflet v3 is shown in the web-eda so my issue not to retrieve a component from v3 was resolved.

From my tests, mapViz mostly worked fine for me and the ref stuffs such as flyTo and thumbnail seemed to function correctly  as well.

bobular:
Thanks to Dave's magic yarn, I've done some quick testing.

For me, the background tile selection is not being screenshotted or remembered in the saved analysis.

moontrip:
> Thanks to Dave's magic yarn, I've done some quick testing.
> 
> For me, a non-standard background tile selection (e.g. satellite) is not being screenshotted or remembered in the saved analysis.

@bobular as far as I know it is not a parameter we keep tracking of, like mouse mode (mouse too) that has a ticket. If you want, I can also try to address it while I am working on the mouse mode stuff https://github.com/VEuPathDB/web-eda/issues/971

moontrip:
@bobular I have extensively tested the case you commented, tile layer thumbnail stuff. Actually what I found is two-fold: first, there exists a bug on the current map (react-leaflet v2) that some tile layers are not working well with the thumbnail/screenshotter . Will figure it out the reason why; b) for v3 (current PR), yes I could confirm that it is somehow not working correctly with tile layers except the default one. Will also look into this.

Thank you for finding the bug which also led to identifying other issue 👍

dmfalke:
> @bobular I have extensively tested the case you commented, tile layer thumbnail stuff. Actually what I found is two-fold: first, there exists a bug on the current map (react-leaflet v2) that some tile layers are not working well with the thumbnail/screenshotter . Will figure it out the reason why; b) for v3 (current PR), yes I could confirm that it is somehow not working correctly with tile layers except the default one. Will also look into this.
> 
> Thank you for finding the bug which also led to identifying other issue +1

Is it possible the thumbnail is generated before the layer has been fully rendered?

moontrip:
@dmfalke Thank you for your comment. FYI, I think that I fixed the issue for the current mapViz that use v2 (not v3). https://github.com/VEuPathDB/web-components/pull/355

I did not look into v3 case yet as I am off today: will check it out next Monday

moontrip:
@bobular bugs are fixed: a) onViewportChange is removed from v3 so it did not work correctly with mapViz. I have added two cases as map events to deal with viewport change, zoomend and moveend. It seems to work fine for me; b) now thumbnail seems to work correctly. Main issue, among others, was that change of baselayer was not recognized, thus it is now configured as another map event.

It would be great if you can test them, when you are available 👍

moontrip:
I have updated this branch with the latest changes at web-components, resolve conflicts, and fix bugs. All stories are working fine from my tests. Noticed some issues with MapViz but they are fixed except one case.

The only issue I could not resolve is that thumbnail is not made for the initial map (after creating new MapViz for the first time): actually it worked fine with screen shoter but with the domToImage it didn't work. If changing zoom, layer, and panning with the initial empty map, then thunbmail is made correctly. Looks like the map/map ref is not ready when creating thumbnail: honestly I have no idea how to deal with this. @dmfalke and/or @jtlong3rd?

dmfalke:
> The only issue I could not resolve is that thumbnail is not made for the initial map (after creating new MapViz for the first time): actually it worked fine with screen shoter but with the domToImage it didn't work. If changing zoom, layer, and panning with the initial empty map, then thunbmail is made correctly. Looks like the map/map ref is not ready when creating thumbnail: honestly I have no idea how to deal with this. @dmfalke and/or @jtlong3rd?

@moontrip we have a similar situation with `PlotlyPlot` where the plot might not be ready when we attempt to make the initial screenshot. You can check https://github.com/VEuPathDB/web-components/blob/640d8f7cf3b2101b49d01a39c36cdfa1def80406/src/plots/PlotlyPlot.tsx for how this is handled. Specifically, see how [sharedPlotCreation](https://github.com/VEuPathDB/web-components/blob/640d8f7cf3b2101b49d01a39c36cdfa1def80406/src/plots/PlotlyPlot.tsx#L115-L119) is used. I think a similar approach will work for MapVEuMap.

moontrip:
> > The only issue I could not resolve is that thumbnail is not made for the initial map (after creating new MapViz for the first time): actually it worked fine with screen shoter but with the domToImage it didn't work. If changing zoom, layer, and panning with the initial empty map, then thunbmail is made correctly. Looks like the map/map ref is not ready when creating thumbnail: honestly I have no idea how to deal with this. @dmfalke and/or @jtlong3rd?
> 
> @moontrip we have a similar situation with `PlotlyPlot` where the plot might not be ready when we attempt to make the initial screenshot. You can check https://github.com/VEuPathDB/web-components/blob/640d8f7cf3b2101b49d01a39c36cdfa1def80406/src/plots/PlotlyPlot.tsx for how this is handled. Specifically, see how [sharedPlotCreation](https://github.com/VEuPathDB/web-components/blob/640d8f7cf3b2101b49d01a39c36cdfa1def80406/src/plots/PlotlyPlot.tsx#L115-L119) is used. I think a similar approach will work for MapVEuMap.

@dmfalke Thank you for your suggestion. Actually I thought of that too but not quite sure if it would work for map case. I will check it out 👍

dmfalke:
I'm happy to take a whack. Just let me know.

moontrip:
> I'm happy to take a whack. Just let me know.

Thanks @dmfalke! 👍 It would be great if you can take a look at it. For your convenience, I made a draft PR implementing sharedPlotCreation used in PlotlyPlot component, but did not work. Please see my comment in there too, which may be helpful for you.
https://github.com/VEuPathDB/web-components/pull/379

moontrip:
updated this branch with the latest changes, e.g., mouse tool work, etc.

moontrip:
As noted in https://github.com/VEuPathDB/web-components/pull/379, @dmfalke resolved the issue of initial thumbnail so the PR is merged into this PR for consistency

moontrip:
updated this react-leaflet-v3: resolved conflicts, fixed bugs, etc

bobular:
Hi @moontrip - I tried to do some testing in web-eda with relative-deps.  Is there a corresponding web-eda branch or PR? I can't find one.

The only way I can get the EDA to run is if I manually install the correct versions as (temporary) direct dependencies of `web-eda` after installing `../web-components` with `relative-deps`.

```
  yarn add react-leaflet@3.2.5 react-leaflet-drift-marker@3.0.0
```

Before that I tried various combinations of `yarn cache clean` and nuking `node-modules` with no luck.

Whatever I did, `yarn why react-leaflet` always told me it was hoisting v2.8.0 from @veupathdb#components#react-leaflet

EVEN THOUGH `node_modules/@veupathdb/components/package.json` was asking for 3.2.5 in its dependencies.

As far as I can tell, this is a simple one-step transitive dependency. I can't see anything else asking for react-leaflet@2.x unless it's the relic of `leaflet-drift-marker` which seems to be hanging around (with react-leaflet 3 we use a different npm: `react-leaflet-drift-marker`. `relative-deps` seems to be failing (at least for me)

I even started wondering if our version of `yarn` is too old. It seems to be an old-ish version of "yarn classic". Migration to yarn 2.x or 3.x seems non-trivial. Didn't look to far into that. @dmfalke 

I guess it will work fine once we publish to npm.

Also on the plus side I've done a fair amount of testing (full screen, regular viz, thumbnails) and everything seems to be working great! 

```
$ yarn why react-leaflet
yarn why v1.22.11
[1/4] Why do we have the module "react-leaflet"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "react-leaflet@2.8.0"
info Reasons this module exists
   - "@veupathdb#components" depends on it
   - Hoisted from "@veupathdb#components#react-leaflet"
info Disk size without dependencies: "1.84MB"
info Disk size with unique dependencies: "2.82MB"
info Disk size with transitive dependencies: "2.98MB"
info Number of shared dependencies: 8
```

moontrip:
@bobular thank you for your great interests, tests, etc.! 👍 
Actually there is no specific web-eda PR because there is no change at web-eda part for `react-leaflet-v3`. I have used `npm-pack-here` and it worked just fine without installing any libraries at web-eda side. And `yarn why` also detecst react-leaflet-v3 correctly. If I remember correctly, there has been a circular dependency issue but it was resolved before by either Dave or Jamie.  

Can you please try npm-pack-here instead and see if it work? I will also check if there is a specific setting I have missed.

moontrip:
@bobular so for test purpose, I just tried to redo all steps at web-eda with npm-pack-here, as listed below

- removed `node_modules` and `local_modules` directories
- `git stash` (remove anything I had at web-eda)
- `yarn cache clean`
- `git pull` and `yarn install`
(at this stage, `yarn why react-leaflet` shows v2.8.0 only)

- `yarn build-npm-modules`: at a web-components directory where you have `react-leaflet-v3-ref` branch (please update your local branch, `git pull origin react-leaflet-v3-ref`)
- `npm-pack-here --target /home/dk/github/react-leaflet-v3-test/web-components` (this directory path varies per one's env)
- `yarn add file:local_modules/@veupathdb/components`
- `yarn install --check-files`

After these steps, `yarn why react-leaflet` shows v3.2.5 only. Personally I prefer npm-pack-here as I am often confused with relative-deps behavior.

Hope this would be helpful for you.

bobular:
Thanks @moontrip - those steps work for me perfectly.

I wonder if `yarn install --check-files` is the secret sauce for `relative-deps`? Anyway, I'm not going to try that now.

Just want to repeat that everything seems to be working, so let's merge soon after the b60 release!

moontrip:
@bobular glad it works for you too! :Thank you for your extensive tests 👍

dmfalke:
> I wonder if `yarn install --check-files` is the secret sauce for `relative-deps`? Anyway, I'm not going to try that now.

Unfortunately, that won't work. Jamie made a fork and attempted to add that capability to relatative-deps, but it didn't work. This is part of the motivation to move to a monorepo. It will allow us to centralize dependency version management.

moontrip:
This PR is closed as https://github.com/VEuPathDB/web-components/pull/445 is merged

PR #348, opened 2022-05-10
preliminary discrete gradient legend

moontrip:
This is an attempt to make a discrete/discretized gradient colormap legend, https://github.com/VEuPathDB/web-eda/issues/1040. It is unknown what kinds of props will be used now, thus it is made at web-components with story and pre-existing Ann's PlotGradientLegend: it is debatable whether PlotLegend and PlotGradientLegend should be combined but possibly there would be many different props each other so just leave them separated for now. 

Here is a screenshot

discrete1

### DIFF
--- a/src/components/plotControls/PlotGradientLegend.tsx
+++ b/src/components/plotControls/PlotGradientLegend.tsx
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { range } from 'd3';
 
 import {
   SequentialGradientColorscale,
   DivergingGradientColorscale,
 } from '../../types/plots/addOns';
+// import { useMemo } from '@storybook/addons';
 
 // set props for custom legend function
 export interface PlotLegendGradientProps {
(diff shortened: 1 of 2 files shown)


asizemore:
@moontrip WOW it looks great!!

If it's not too much trouble, can we flip the direction of how the gradient is displayed so that black, 100.00 is at the top and the light gray, 0.00? I realize i made a mistake in [my mockup in the original issue](https://github.com/VEuPathDB/web-eda/issues/1040#issuecomment-1103197915) and had the lightest, lowest numbers at the top, so that was totally my fault I'm sorry!

moontrip:
@asizemore Thank you for your information! Actually I had to flip the array to follow your mockup 😄 So, I can probably do that direction: will do that!

moontrip:
@asizemore As you suggested, I have changed the order to descending order as shown in the following screenshot 👍 

discrete-gradient-descending

asizemore:
Woot thanks @moontrip !! It looks great!

PR #351, opened 2022-05-16
ChartMarkers new props for provided x-axis labels

bobular:


src/map/DonutMarker.tsx
```diff
@@ -127,14 +129,15 @@ export default function DonutMarker(props: DonutMarkerProps) {
...
+  const sumValues: number = fullStat.slices
     .map((o) => o.value)
     .reduce((a, c) => {
       return a + c;
     });
-  //DKDK convert large value with k (e.g., 12345 -> 12k): return original value if less than a criterion
-  let sumLabel: number | string = kFormatter(sumValues);
+
+  // for display, convert large value with k (e.g., 12345 -> 12k): return original value if less than a criterion
+  const sumLabel: number | string = props.markerLabel ?? kFormatter(sumValues);
```
moontrip, line 140: @bobular I have tested your PRs at web-eda and web-components and they seem to work properly. Just one note in this PR, I noticed that markerLabel is string and it is not shortened with "k". 5 digit number is barely fitted to the pie plot but I guess there would be the case larger digits than that. Since it is basically number string, I think you can convert it into number and then pass it to kFormatter function?

Though, not sure if we abandoned the use of "k" for a large number.

bobular: Done!

bobular: I think the bar chart numbers should be comma-formatted for numbers over 2100 

I found some relevant code here
```
web-eda/src/lib/core/components/filter/HistogramFilter.tsx:        useGrouping: !(Number(value) >= 1900 && Number(value) <= 2100),
```
will add this soon.

moontrip: @bobular Honestly I can't remember why I did it before, but your change would work 👍

moontrip:
@bobular since I am commenting on this PR, I will make another one here :) So, it seems like you have decided to add comma for the barplot, it may cause two potential issues. 

a) donut marker also shows comma, instead of "k". Seems like the format of the number label may need to be handled at mapViz?
map-pieplot-k1

b) for barplot, it may be an issue for number like hundred thousands, 123456, which may go beyond the outline? ten thousands are barely fitted into the boundary for now: maybe just ok with your test as you also introduced million "M".
map-bar-proportion-k1

bobular:
Thanks @moontrip - good catch for the loss of the 'k' in the donuts, I will look into that.

In FF six digit numbers with a comma do fit under the charts.  Just...

bobular:
I have addressed this now, in both PRs.

dmfalke (approved):

moontrip (approved):

PR #353, opened 2022-05-18
Ensure east-west bounds are slightly different when fully zoomed out

### LINKED ISSUE #352: Map markers disappear at minimum zoom level
When fully zoomed out, markers disappear in the map visualization (SCORE Seasonal, SCORE Mozambique, UMSP)

Presumably a problem with the longitude bounds being more than a "whole world" but will investigate.

bobular:
Resolves #352 

Doesn't need two reviewers but thought you both might be interested.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.16.1",
+  "version": "0.16.2",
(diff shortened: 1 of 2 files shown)


moontrip (approved):
I have tested this and it works out 👍

PR #354, opened 2022-05-19
remembering the selection of mouse tool

moontrip:
This addresses https://github.com/VEuPathDB/web-eda/issues/971 with corresponding web-eda PR, https://github.com/VEuPathDB/web-eda/pull/1141.

Map related stories are updated as well.

### DIFF
--- a/src/map/MapVEuMap.tsx
+++ b/src/map/MapVEuMap.tsx
@@ -163,6 +163,10 @@ export interface MapVEuMapProps {
   showNoDataOverlay?: boolean;
   /** Whether to show the Scale in the map */
   showScale?: boolean;
+  /** mouseMode control */
+  mouseMode: 'default' | 'magnification';
+  /** a function for changing mouseMode */
+  setMouseMode: (value: 'default' | 'magnification') => void;
 }
 
 function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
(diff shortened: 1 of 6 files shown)


bobular (changes requested):
I was going to suggest that you export a MouseModes type from MouseTools, but I see you already do!!

Please could you import that into the Stories (and perhaps also export a `const defaultMouseMode : MouseMode = 'default'` so you can say

```
  const [mouseMode, setMouseMode] = useState<MouseMode>(defaultMouseMode);
```
?

This would make it more future-proof to the addition of more modes!


I also recommend changing the `setMouseMode` property to `onMouseModeChange` to be more consistent (though I see we are not very consistent with the use of `Changed` and `Change`)
I notice on the EDA side you have called it `onMouseModeChange`

moontrip:
@bobular Thank you for your review and suggestions! 👍 I have addressed your feedbacks.

src/map/MapVEuMap.tsx
```diff
@@ -156,6 +156,10 @@ export interface MapVEuMapProps {
   showNoDataOverlay?: boolean;
   /** Whether to show the Scale in the map */
   showScale?: boolean;
+  /** mouseMode control */
+  mouseMode: 'default' | 'magnification';
```
bobular, line 160: We could save some typing/maintenance with
```suggestion
  mouseMode: MouseMode;
```

src/map/MapVEuMap.tsx
```diff
@@ -156,6 +156,10 @@ export interface MapVEuMapProps {
...
+  /** a function for changing mouseMode */
+  onMouseModeChange: (value: 'default' | 'magnification') => void;
```
bobular, line 162:
```suggestion
  onMouseModeChange: (value: MouseMode) => void;
```

bobular (approved):
Looks great, but see my suggestions :-)

PR #355, opened 2022-05-19
replacing some baselayers for making thumbnails correctly

moontrip:
This addresses https://github.com/VEuPathDB/web-eda/issues/1142. 

TL;DR: I recommend to switch two layers to another ones as temporarily added to the existing tilelayers, named as Terrain_Alt1 and Light_Alt1.

I have dug several parts of functions to figure out the reason why some tilelayers did not work for making thumbnails correctly, however, in a code wise, I could not spot on something special.

After testing more extensively, I figured it out that only two tilelayers such as Terrain and Light did not work out: the latter sometimes worked but was not that consistent. I recalled myself that the two titlelayers were relatively new ones to be added because previous ones seemed to be no longer supported. Those two come from the same provider, and I suspect that they are like an overlay style layer on top of OpenStreetMap layer: though I could not find that info, just my pure guess. Whenever making a thumbnail from one of the two tilelayers, it shows OpenStreetMap layer, not their own layer.

Accordingly, I tried to find alternatives for those two layers, which worked just fine for thumbnails without specific issue. FYI, below are screenshots of alternative tilelayers. If those are acceptable, then I will replace with them.

- Terrain alternative (more like topographic map)
terrain2

- Light alternative
light2

### DIFF
--- a/src/map/MapVEuMap.tsx
+++ b/src/map/MapVEuMap.tsx
@@ -97,6 +97,18 @@ export const baseLayers = {
     // maxZoom='18'
     // noWrap='0'
   },
+  Terrain_Alt1: {
+    url: 'https://{s}.tile.opentopomap.org/{z}/{x}/{y}.png',
+    attribution:
+      'Map data: &copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors, <a href="http://viewfinderpanoramas.org">SRTM</a> | Map style: &copy; <a href="https://opentopomap.org">OpenTopoMap</a> (<a href="https://creativecommons.org/licenses/by-sa/3.0/">CC-BY-SA</a>)',
+    maxZoom: 17,
+  },
+  Light_Alt1: {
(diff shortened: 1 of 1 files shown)


dmfalke:
Fascinating!

bobular:
Hi @moontrip - which layers are not working in the thumbnails?  In FF they are all working for me.

bobular:
Oh, I see it's very patchy in Chrome...

moontrip:
@bobular Thank you for your tests! Honestly, I didn't know it worked fine at Firefox: in most cases I use Chrome. 
As shown in your screenshot, that's the reason why I said inconsistent behavior in the mtg.

dmfalke:
Is there any way to confirm that the screenshot is taken _after_ all parts of the layer have been loaded and rendered by the browser? It almost looks like it is taken before the layer has finished loading.

moontrip:
@dmfalke I think it might be the reason why Connor did set a timeout (1000 ms) when he implemented the thumbnail capture (screenshotter), allowing some time.

bobular:
Closing after discussion in EDA dev mtg

PR #356, opened 2022-05-24
making log scales for both X and Y in the Scatter plot

moontrip:
This makes log scale controls for both X and Y axes in the Scatter plot and its corresponding Viz. https://github.com/VEuPathDB/EdaNewIssues/issues/292

Since Scatter plot (and Viz) allow date variable at X and Y axes, it is configured to disable the log scale control when the selected variable is date type.

Here are example screenshots with explanation in the title: the same study and variables with the screenshot in the ticket were used for a comparison purpose.

- log scale is on for X-axis
scatter-logscale1

- controlling legend to show/hide data: this may be useful for scattered data that show large range of confidence intervals in Y-axis.
scatter-logscale2

- A demo to show that log scale control at the Y-axis is disabled because its variable type is date
scatter-logscale-date1

### DIFF
--- a/src/plots/ScatterPlot.tsx
+++ b/src/plots/ScatterPlot.tsx
@@ -6,6 +6,10 @@ import {
   OrientationAddon,
   OrientationDefault,
   AxisTruncationAddon,
+  independentAxisLogScaleAddon,
+  independentAxisLogScaleDefault,
+  DependentAxisLogScaleAddon,
+  DependentAxisLogScaleDefault,
 } from '../types/plots';
 // add Shape for truncation
 import { Layout, Shape } from 'plotly.js';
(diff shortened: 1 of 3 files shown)


src/plots/ScatterPlot.tsx
```diff
@@ -117,27 +125,54 @@ const ScatterPlot = makePlotlyPlotComponent(
...
               extendedIndependentAxisRange?.min,
               extendedIndependentAxisRange?.max,
-            ]
+            ].map((val) =>
+              independentAxisLogScale && val != null
+                ? val <= 0
+                  ? -0.1
+                  : Math.log10(val as number)
+                : val
+            )
```
dmfalke, line 134: I'm a little confused by this. If the axis is being scaled, why do we need to scale the data, too? Is it related to how we're doing axes (in a different trace, maybe)?

dmfalke (approved):

moontrip:
@dmfalke Actually, that's quite a confusing part when using plotly.js for a log scale. Plotly treats log scale case as log data, even if the axis tick labels are shown as actual numbers. Strange enough but perhaps it is because such a treatment may be easier to figure out the numbers visually, when looking at the plot. I am always confused about that concept 😄

Issue #330, opened 2022-03-24
When scrolling down a page with a map, do not zoom map when the pointer goes over it


bobular:
This can be demoed in the Map->General->Windowed story if you have a small enough browser window.

I think the zooming is an undesired behaviour.  Not sure of priority of this ticket.  Solution may be low cost but who knows!

Many sites require ctrl-scroll to zoom, but I think we could try the following solution:

https://gis.stackexchange.com/questions/111887/leaflet-mouse-wheel-zoom-only-after-click-on-map

moontrip:
@bobular I have tested this and got mixed results. Basically it works as is: when using mouse-wheel scrolling in the outside of map (at left and right side of the map), it scrolls the page, but if scrolling inside the map, it works as zoom in and out. So, it is intuitively working, I suppose.

Though, with several tests with both Chrome and Firefox browsers, I found one exception at Firefox. When scrolling down from just above the (windowed) map (outside the map though), the page moves down but at the same time, map also zooms out. However, this phenomenon does not happen at Chrome browser. I am not quite sure if there is a solution, as it seems to be specific to Firefox only. 

I have also tried to use a method you linked above but it did not help.

bobular:
@moontrip - aha, I see. The Chrome behaviour is basically what we want, if low cost enough to implement...

dmfalke:
fwiw, on gene record pages, we have a checkbox to toggle the ability to scroll/zoom in an embedded jbrowse window:
![image](https://user-images.githubusercontent.com/365139/169558208-eab12bba-1250-415d-b0ca-f1627cb2f20f.png)

It's worth considering for consistency.

danicahelb:
@moontrip i moved this out of "this cycle" as it is only a nice to have and we are in crunch time for b58. thanks!

PR #357, opened 2022-05-31
add scroll and zoom option to map

moontrip:
@bobular and @dmfalke

This is an implementation to add a checkbox of scroll and zoom as suggested by Dave which is related to this issue,  https://github.com/VEuPathDB/web-components/issues/330

As a test and demo purpose, this is made as a draft PR. A story can be found at Map -> General -> Scroll and Zoom for testing. 

scrollAndZoom1


src/map/MapVEuMap.tsx
```diff
@@ -163,6 +163,8 @@ export interface MapVEuMapProps {
   showNoDataOverlay?: boolean;
   /** Whether to show the Scale in the map */
   showScale?: boolean;
+  /** is map scroll and zoom allowed? */
+  isMapScroll?: boolean;
```
dmfalke, line 167: I'm not sure this is the clearest name for this property. What about something like `scrollingEnabled`?

dmfalke (approved):
This looks good to me!

src/map/MapVEuMap.tsx
moontrip: thank you for your review and suggestion, @dmfalke. Yep it would work 👍 Will change

moontrip: @dmfalke changed as suggested 👍

moontrip:
@bobular @dmfalke I wonder if we need this or just leave the map as is

bobular:
Sorry I didn't see this as I wasn't a reviewer!
I will have a play tomorrow.

moontrip:
sorry @bobular At that time I made this draft PR, I just wanted to make sure whether my implementation was what Dave wanted. I just added you as a reviewer too

src/map/MapVEuMap.tsx
```diff
@@ -186,6 +188,7 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
     showSpinner,
     showNoDataOverlay,
     showScale = true,
+    scrollingEnabled,
```
bobular, line 191: Hi @moontrip - please could you make this default to true, then I would like to approve and merge this PR and use the new prop for the FSM minimap.
I might do this before you log on for the day. Let's keep in touch via Slack.
```suggestion
    scrollingEnabled = true,
```

bobular: Actually, I think I was trying to kill two birds with one stone.

Disabling scroll-wheel functionality isn't enough for the minimap.

I'll add an `interactive` (defaults to true) prop for MapVEuMap in my fsm-minimap branch.

Still happy for this scrollingEnabled prop to be added though :-)

moontrip: @bobular this took some time to resolve merge conflicts, but I have applied your suggestion to the lasted commit 👍

bobular (approved):
Cool, thanks!

PR #361, opened 2022-06-07
adding increased margin for scatterplot log scale

moontrip:
This addresses a suggestion, https://github.com/VEuPathDB/web-eda/issues/1153#issuecomment-1142637556, which needs to increase margin for scatterplot log scale

After testing several cases, one of the simplest solution is to increase the margin when log scale is used. Screenshot is added for a comparison with Danica's screenshot at the comment above. As shown, marker is not cut in this implementation.

Note that this is only applied to max range of dependent axis because it would be most likely the case we need to consider.

scatter-logscale-margin

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.16.2",
+  "version": "0.16.3",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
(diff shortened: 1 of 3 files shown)


jtlong3rd (approved):

moontrip:
@jtlong3rd Oh! as usual, you are so fast! 👍 ⚡ (lightening, I mean 😄)

PR #366, opened 2022-06-15
Map component: cancel debounce and use the debounced version for the 'one off' call too

### LINKED ISSUE #362: Map has a runtime exception when closing a visualisation (in EDA) within 1 second of zooming or panning
Make a map viz in EDA. Pan the map a bit, then close it (or click "Visualize" tab) within 1 second of moving the map.

I can't seem to find a similar way to reproduce in Storybook.

I think this is due to an uncancelled debounced function `updateBounds` in SemanticMarkers.tsx

Should be easy enough to fix. Please take it if you fancy it. I will be doing another ticket first.

bobular:
Resolves #362 

I have tested this and it solves the 1 second exception, and changing `updateBounds()` to `debouncedUpdateBounds()`  doesn't cause any noticeable delay or increase the number of back-end requests.

@jernestmyers @moontrip and @chowington may be interested to see this small change (though I think I originally wrote the buggy code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.16.5",
+  "version": "0.16.6",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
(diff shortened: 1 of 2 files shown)


moontrip:
@bobular I have quickly tested this. It was very hard for me to replicate the symptom even before applying this PR, but I could eventually see the case. After applying this PR, it seems to work fine.

bobular:
> @bobular I have quickly tested this. It was very hard for me to replicate the symptom even before applying this PR, but I could eventually see the case. After applying this PR, it seems to work fine.

Thanks DK - you have to click away from the viz quickly!

PR #368, opened 2022-06-17
Use dom-to-image directly

dmfalke:
This PR uses `dom-to-image` directly for generating thumbnails. It appears to fix the freezing issues in Chrome. I also update the light and terrain layers so that they are using CORS-friendly domains.

related to https://github.com/VEuPathDB/web-eda/issues/1118


moontrip (approved):
The new dome-to-image worked fine: even with the setTimeout of 1000 ms at thumbnails.ts. New layers are also working fine for me.

So, I think that we can use the reduced 1000 ms setTimeout at web-eda's thumbnails.ts

bobular (approved):
Nice! No freezing in Chrome, still works in FF :-)

dmfalke:
I commented out the light and dark layers. We can revisit later. I will raise this on slack

moontrip (approved):

bobular (approved):
Still works great in Chrome.

I'm cool with removing the two layers, but it does mean that some users' saved visualization configs don't parse on opening the thumbnail. This is because the `vizConfig.baseLayer` type is imported from `web-components/MapVEuMap`.  The user just sees a default empty visualization.  I think this is probably OK.  Backwards compatibility is definitely something to put on the tech debt list!

```
const MapConfig = t.intersection([
  t.type({
    mapCenterAndZoom: t.type({
      latitude: t.number,
      longitude: t.number,
      zoomLevel: t.number,
    }),
    baseLayer: t.keyof(baseLayers),
  }),
```

PR #369, opened 2022-06-20
Experimental tiny map 

bobular:
Quick hacky solution (but have added several new MapVEuMap props) to make a new story Map-->General-->Tiny

Which makes a tiny donut map with small markers, no numbers, and a finer-grained geohash mapping.

Could be part of the entity diagram in MapVEu-is-EDA.  I used it to create this [mock-up](https://docs.google.com/presentation/d/17Yg4Kz00pW6wbNo3bK0DpQjW4cVEzrq0VQ75pw1xZNc/edit#slide=id.g1333a64e1ab_0_5):

image


bobular:
Note that the main map will need `minZoom: 0`
It seems that the whole world will just about fit into 220/110 pixels...
Hopefully that doesn't cause any problems.
image

bobular:
Closing this PR and branching to `fsm-minimap` for futher work

PR #363, opened 2022-06-11
adding log scale to lineplot

moontrip:
This is related to https://github.com/VEuPathDB/web-eda/issues/1157.

In this commit, it is for adding log scale to lineplot

PR #371, opened 2022-06-30
Viz log range truncation

moontrip:
This is branched out from https://github.com/VEuPathDB/web-components/pull/363 to address two things for scatter and line plots:

a) set log scale range correctly
b) truncation

This is related to the web-eda PR, https://github.com/VEuPathDB/web-eda/pull/1241

To reviewers, @bobular & @asizemore: 
As mentioned above, I have branched out from https://github.com/VEuPathDB/web-components/pull/363 to address the two tasks. I have separated them because this one would have taken some time while you were reviewing the previous PR. 

Now, you can review this PR instead of https://github.com/VEuPathDB/web-eda/pull/1241.

### DIFF
--- a/src/plots/LinePlot.tsx
+++ b/src/plots/LinePlot.tsx
@@ -10,6 +10,10 @@ import {
   OrientationAddon,
   OrientationDefault,
   AxisTruncationAddon,
+  independentAxisLogScaleAddon,
+  independentAxisLogScaleDefault,
+  DependentAxisLogScaleAddon,
+  DependentAxisLogScaleDefault,
 } from '../types/plots';
 // import truncation util functions
 import { extendAxisRangeForTruncations } from '../utils/extended-axis-range-truncations';
@@ -22,6 +26,8 @@ export interface LinePlotProps
   extends PlotProps<LinePlotData>,
     // truncation
     OrientationAddon,
+    independentAxisLogScaleAddon,
+    DependentAxisLogScaleAddon,
     AxisTruncationAddon {
   /** x-axis range: required for confidence interval - not really */
   independentAxisRange?: NumberOrDateRange;
@@ -64,6 +70,8 @@ const LinePlot = makePlotlyPlotComponent('LinePlot', (props: LinePlotProps) => {
     // add axis range control truncation
     orientation = OrientationDefault,
     axisTruncationConfig,
+    independentAxisLogScale = independentAxisLogScaleDefault,
+    dependentAxisLogScale = DependentAxisLogScaleDefault,
     ...restProps
   } = props;
 
@@ -80,7 +88,10 @@ const LinePlot = makePlotlyPlotComponent('LinePlot', (props: LinePlotProps) => {
   const extendedDependentAxisRange = extendAxisRangeForTruncations(
     standardDependentAxisRange,
     axisTruncationConfig?.dependentAxis,
-    dependentValueType === 'date' ? 'date' : 'number'
+    dependentValueType === 'date' ? 'date' : 'number',
+    // adjust range for log scale
+    'lineplot',
+    dependentAxisLogScale
   );
 
   // make rectangular layout shapes for truncated axis/missing data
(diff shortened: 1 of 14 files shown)


asizemore:
Thanks @moontrip ! I have this on my todo list for this afternoon :)

asizemore:
Would we be able to force at least two y-axis labels to show up? In India ICEMR Meghalaya Cross-sectional (no filters) i can make the following plot

Screen Shot 2022-06-30 at 4 59 23 PM

src/plots/LinePlot.tsx
```diff
@@ -10,6 +10,10 @@ import {
   OrientationAddon,
   OrientationDefault,
   AxisTruncationAddon,
+  independentAxisLogScaleAddon,
+  independentAxisLogScaleDefault,
+  DependentAxisLogScaleAddon,
```
bobular, line 15: I just noticed the inconsistent lowercase `i` and capital `D` here!

Maybe the Addon should be capitalised and the default value start with lowercase?

Though the OrientationAddon/Default are both uppercase. I guess go with uppercase!

bobular (commented):
I'll do a proper review including the code tomorrow. 

I just noticed that there seems to always be one ChartMarker (on map viz, bar plot (count) mode) that doesn't change when toggling log mode.  It seems fine in proportion mode. I'll investigate more tomorrow.

bobular:
> I'll do a proper review including the code tomorrow.
> 
> I just noticed that there seems to always be one ChartMarker (on map viz, bar plot (count) mode) that doesn't change when toggling log mode. It seems fine in proportion mode. I'll investigate more tomorrow.

Investigating...

The magnified mouse-over plot IS changing between log/no log mode.

It seems to only affect the chartmarker that contains the tallest bar (e.g. goes up to the maximum of the y-range)

The log-scale storybook story chartmarkers don't seem to respond to the toggle switch at all (but the mouse over magnifications do).

I will keep investigating!

bobular:
> I will keep investigating!

Hi @moontrip - I think I have solved the conundrum...

I think the issue is that the SVG chartmarker code doesn't currently support log scale. This is why in EDA the one marker with the maximum bar height doesn't change when you toggle.  In storybook I'm not quite sure what's going on with y-ranges, but it must be a similar issue.

moontrip:
@bobular First of all, thank you for taking over this PR! 👍 I have looked at your changes and tested them, and they looked good to me! Since I am the author of this PR, I can't review it myself but I approve this PR anyway.  😄

src/utils/extended-axis-range-truncations.ts
```diff
@@ -19,8 +19,8 @@ export function extendAxisRangeForTruncations(
   // set this to avoid error
   if (axisRange == null) return undefined;
 
-  // adjust margin per log scale
-  const noTruncationMargin = logScale ? 0.3 : 0.02;
+  const truncationMarginFactor = 0.05; // how much padding needed for the yellow truncation warning
+  const noTruncationMarginFactor = 0.02; // how much padding needed for, e.g. scatter plot points, even when not truncated
```
asizemore, line 23: is this meant to replace 0.02 on line 43?

src/utils/extended-axis-range-truncations.ts
```diff
@@ -75,18 +75,31 @@ export function extendAxisRangeForTruncations(
     } else {
       // consider padding
       const diff = (axisRange.max as number) - (axisRange.min as number);
+      const ratio = (axisRange.max as number) / (axisRange.min as number);
       const axisLowerExtensionStart = config?.min
-        ? (axisRange.min as number) - 0.05 * diff
+        ? logScale && isFinite(ratio) && ratio > 0
+          ? (axisRange.min as number) /
+            10 ** (truncationMarginFactor * Math.log10(ratio))
```
asizemore, line 82: @bobular mind walking me through the math a bit? I would have guessed something like log(axisRange.min) - log(truncationMarginFactor) would be sufficient?

src/utils/logscale-dtick.ts
```diff
@@ -0,0 +1,28 @@
+import { NumberOrDateRange } from '../types/general';
+import { isNumber } from '../types/guards';
+
+/**
+ * Given an axis range, this will usually a value of 1 to be used in Plotly's layout.{x,y}axis.dtick,
+ * which indicates 10-fold tickmark spacing (it's calculated as 10**dtick)
+ *
+ * However, if the ratio between range.max/range.min is less than 10, we need to use a smaller spacing
+ * so that we get some tick marks.
```
asizemore, line 9: very cool!

asizemore (commented):
Code looked good, just had a few questions. Also i didn't see any changes in Chart Markers > Log Scale when i toggle on/off Log Scale in storybook. What should I be looking for here?

Over to web-eda!

bobular:
Just a quick email reply about chart markers - we have that in another
ticket now. DK is on it 👍

dmfalke (commented):
Overall, I think this looks good. The one suggestion I have right now involves. https://github.com/VEuPathDB/web-components/blob/31f001ff771e01811efafde2385743fceec2e3a8/src/utils/extended-axis-range-truncations.ts#L16.

Instead of passing in the plotType, would it be better to change this parameter to something like `addPadding: boolean = false`? That way we aren't proliferating the untyped plotType string.

src/utils/extended-axis-range-truncations.ts
bobular: So we have to include ratio (= max/min) somehow.

log10(ratio) gives us the number of (base 10) ticks on the axis.  e.g. log10(1000/0.1) = 5, ticks at 0.1, 1, 10, 100, 1000

So in log-space the length of the axis is 5.  We calculate a fraction of that to add as the margin, but have to convert it back into "regular-space" with the 10^

Just to be more confusing, just before the axis range gets passed to Plotly, we log the range's min+max when log-scale is active.

bobular (approved):
I've done the thing Dave suggestion with the `addPadding` arg and am now approving!

dmfalke (approved):

PR #372, opened 2022-07-05
Address modal bug and tweak styling in faceted plots

jernestmyers:
Resolves [this issue](https://github.com/VEuPathDB/EdaNewIssues/issues/356).

Added a guard to ensure the modal isn't rendered at an inappropriate time- which was the root of the bug that Bob discovered. Also tweaked the styles as shown in the issue's comments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.16.7",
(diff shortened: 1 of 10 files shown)


bobular (approved):
Tested as hard as I could and looks good!

Issue #380, opened 2022-07-29
Fix map chart markers SVG code to handle logscale prop


bobular:
To follow the latest work on viz axes, ranges, logscale and truncation: https://github.com/VEuPathDB/web-eda/pull/1241

Maps in SCORE Mozambique (for example)

The mouse-over bar charts are correctly showing the log scale. The SVG mini chart markers need to take into account the log transformation.

moontrip:
@bobular I tried to address this but I could not make an appropriate solution because there seems no way to handle log scale using the pure SVG. This would probably need to use a graph lib like Plotly etc. to address. However, as we observed with Connor's histrogram approach for the chart marker in the beginning of the react map work, its use would make map slower, which was the reason why we used SVG chart marker instead.

In an attempt to address this in a simpler way with some degree of error, I made two draft PRs at web-eda and web-components respectively, https://github.com/VEuPathDB/web-eda/pull/1290 & https://github.com/VEuPathDB/web-components/pull/381. It may be used to minimize a visual gap between small (marker) and large (using mouse tool's magnification) charts for the time being, if necessary: it is more accurate for count option (plot mode) than proportion option, I suppose.

bobular:
Hi @moontrip - one problem is that the current code does not honour `dependentAxisRange.min` properly.  I am attaching some sketches I made for the calculation of `barHeight`.  `R_min` and `R_max` are the values of `dependentAxisRange`.  Note that the calculations are basically the same in both cases, but with logScale active, the value, and `R_mn` and `R_max` are all logged.

It would be great also to get rid of the copy-paste duplication resulting from the `if (globalMaxValue) ... else `. It looks to me identical except for in one case there's a global axis range, and in the other a local one.

![00-no-logs](https://user-images.githubusercontent.com/308639/182680241-6bc52e90-0b1e-4d94-af45-147104fd1a6e.jpg)
![01-with-logs](https://user-images.githubusercontent.com/308639/182680234-217a78f4-2ccf-4fa8-8c18-212d93c1c50e.jpg)

moontrip:
@bobular Thank you for taking time to think about this. Most of all, we do not have axis range control for mapViz. 😄 I have checked that we only have two kinds of controls for mapViz, plot mode and y-axis log scale (under the branch of viz-log-range-truncation)

bobular:
That's correct there is no current y axis range control for mapviz, so the
`range.min` is always zero or I think usually 0.1 in logscale mode
[...]

PR #381, opened 2022-08-02
addressing SVG log scale

moontrip:
This draft PR is branched out from https://github.com/VEuPathDB/web-components/pull/371 as this needs map's log scale functionality

This is a temporary solution, just minimizing gap between small and large (using mouse tool) charts: Please see this comment for more detailed info: https://github.com/VEuPathDB/web-components/issues/380#issuecomment-1203128901.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.17.1",
+  "version": "0.17.2",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
--- a/src/map/ChartMarker.tsx
+++ b/src/map/ChartMarker.tsx
@@ -90,12 +90,12 @@ export default function ChartMarker(props: ChartMarkerProps) {
         });
 
   // find local min/max (one marker)
-  const localMinValue: number = Math.min(
-    ...fullStat.map((o) => o.value).filter((a) => a > 0)
-  ); // max of fullStat.value per marker icon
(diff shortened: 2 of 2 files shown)


moontrip:
updated as noted at https://github.com/VEuPathDB/web-components/issues/380#issuecomment-1205613883

moontrip:
@bobular it would be great if you can test this PR when you will be available

bobular (approved):
This is looking and working great.

I only have one optional suggestion: the localMin and localMax values are calculated even if they are not needed. I know it's not very expensive, but you know me! Perhaps it could be calculated conditionally, something like

```
  const yAxisRange = props.dependentAxisRange
    ? props.dependentAxisRange
    : { min: Math.min(...), max: Math.max(...) };
```
(probably with a `as NumberRange` ;-) )

and use yAxisRange in the calculations of barHeight (which also simplifies the code there, I think)

moontrip:
@bobular I have addressed your suggestions by introducing localMinMaxPosRange that includes min/max computation conditionally. 👍

moontrip:
@bobular I have changed this PR with the discussions we had. It looks fine for me but it would be great if you can test this too 👍

bobular:
> @bobular I have changed this PR with the discussions we had. It looks fine for me but it would be great if you can test this too +1

Hi @moontrip - I haven't tested it but I looked at the code carefully and it gets the :+1:  from me!

moontrip:
Thank you for your review, @bobular 👍

Issue #344, opened 2022-04-14
Widen viz plot legend to 400px and take care of ellipsis processing too


bobular:
With the new layout going out in b57, there is space available to widen the legend to 400px (used by the coverage table and birds eye plot) - with 10px padding I think.

The length threshold for the ellipsis handling may also need changing. Ideally this should be a prop, not hardcoded.

Looks like the code is in PlotLegend.tsx

While there we should see if we can remove some EDA-specific behaviour (from the comment below, at least)
```
            maxHeight: 250, // same height with Scatterplot R-square table
```

dmfalke:
The ellipsis behavior can be controlled via the CSS rule `text-overflow: ellipsis`. This will allow things to be more dynamic and responsive. See https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow.

I think the width should probably be a prop, or controlled by the component rendering this one.

bobular:
True, and I have even used it recently, duh!
I think the character-based truncation was a hangover from the days when we had to hack Plotly's tooltips.

bobular:
While we are at it, I think the "title" of the legend needs some formatting tweaks.
Maybe it just needs to be in bold, or perhaps the spacing is a bit off.  Certainly for this very short "Sex" one below it doesn't look quite right.

![image](https://user-images.githubusercontent.com/308639/163481335-f2876a6b-3f9a-42cf-9a7b-552082d3f8bd.png)

SheenaTomko:
Looks good. Legend box is now wider by default and the variable name within it is bolded

PR #382, opened 2022-08-03
changing legend styles

moontrip:
This will address https://github.com/VEuPathDB/web-components/issues/344

Done based on several comments by Bob and Dave in the ticket
- [x] set legend width to be 400px like other tables (plz see below screenshot)
- [x] introduced `text-overflow: ellipsis` instead of a functional approach: removed the function, `legendEllipsis` 
- [x] set bold for legend title to distinguish from legend labels
- [x] made some left and bottom margins for better display

For reference, a screenshot is attached in the following:
legend-size-adjust1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.17.0",
+  "version": "0.17.1",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
--- a/src/components/plotControls/PlotLegend.tsx
+++ b/src/components/plotControls/PlotLegend.tsx
@@ -17,7 +17,6 @@ interface PlotLegendProps {
   legendItems: LegendItemsProps[];
   checkedLegendItems: string[] | undefined;
   legendTitle?: string;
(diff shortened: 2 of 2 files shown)


moontrip:
made another screenshot to be compared with the previous example in the ticket

- before (taken from the ticket)
163481335-f2876a6b-3f9a-42cf-9a7b-552082d3f8bd

- after (this PR)
histogram-legend1

bobular (approved):
All looks great. Confirm the ellipsis mouse-overs are working fine for the legend values.

I couldn't find a variable name long enough to test the legend title, but it's possible that this also needs `textOverflow: ellipsis` ?

bobular:
> All looks great. Confirm the ellipsis mouse-overs are working fine for the legend values.
> 
> I couldn't find a variable name long enough to test the legend title, but it's possible that this also needs `textOverflow: ellipsis` ?

Ah, I found one - the title is wrapped, which I think is the better solution!

image

moontrip:
> 

Thank you for your review and tests, @bobular 👍 Actually I didn't think of ellipsis on the legend title as it is quite rare to see very long title, especially with the increased width and wrapping. Glad you like the wrapped title 😃

PR #383, opened 2022-08-08
Use CoreUI switch

chowington:
Works toward https://github.com/VEuPathDB/web-eda/issues/1252
Requires https://github.com/VEuPathDB/CoreUI/pull/101


jernestmyers (approved):
Looks good, though I'll admit it was unclear to me why we were passing in arrays for the `styleOverrides` to the `default` and `hover` states. I had to look over the `FilledSwitch` story to make sense of why, which makes me wonder if we should add some documentation in the CoreUI `SwitchStyleSpec` type definitions?

asizemore (commented):
I don't quite understand - why would we wrap `Toggle` in `Switch`? Are we gaining anything that we shouldn't add to `Toggle`?

chowington:
> I don't quite understand - why would we wrap `Toggle` in `Switch`? Are we gaining anything that we shouldn't add to `Toggle`?

Looks like the only thing we really get from it is the single color prop, which Toggle doesn't actually listen to currently. See this comment for further thoughts on this: https://github.com/VEuPathDB/CoreUI/pull/101#discussion_r962071694

PR #384, opened 2022-08-10
power format of axis label

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1275

I have added the power format for Viz/axis who has log scale control such as Histogram Y-axis, barplot Y-axis, Scatterplot X and Y axes, and Lineplot X and Y axes.

For Histogram and Barplot, it was found that they had tickformat attribute that formats large number, so it is removed to show power format. 

Just for an information, Plotly.js actually supports minexponent attribute which controls the starting point of power format display. However, @types/plotly.js somehow do not have type for that. I saw that Danica wanted to show power format from five-figure (e.g., 10^4), which is the default value of minexponent (minexponent: 3), thus I did not consider typing for that: honestly, I could not find a way to extend type to support minexponent without changing @types/plotly.js either. 

Here is a screenshot with power format for the example case shown in the ticket.
scatter-axis-power-format

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.17.4",
+  "version": "0.17.5",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -203,7 +203,6 @@ const Barplot = makePlotlyPlotComponent(
 
     const dependentAxisLayout: Layout['yaxis'] | Layout['xaxis'] = {
       automargin: true,
-      tickformat: dependentAxisLogScale ? ',.1r' : undefined, // comma-separated thousands, rounded to 1 significant digit
       hoverformat: dependentAxisLogScale
         ? dataLooksFractional
           ? ',.4f'
@@ -229,6 +228,8 @@ const Barplot = makePlotlyPlotComponent(
         ? logScaleDtick(extendedDependentAxisRange)
         : undefined,
       showticklabels: showDependentAxisTickLabel,
+      showexponent: 'all',
+      exponentformat: 'power',
     };
 
     const layout: Partial<Layout> = {
--- a/src/plots/Barplot.tsx
+++ b/src/plots/Barplot.tsx
@@ -23,7 +23,7 @@ import { NumberRange } from '../types/general';
 // import truncation util functions
 import { extendAxisRangeForTruncations } from '../utils/extended-axis-range-truncations';
 import { truncationLayoutShapes } from '../utils/truncation-layout-shapes';
-import { logScaleDtick } from '../utils/logscale-dtick';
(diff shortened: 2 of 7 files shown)


bobular (commented):
I managed to reproduce an old bug similar (but not identical) to what we used to get before I introduced the `tickformat: ',.1r'` for some plot types

image

Here is how it looks on qa

image

I realise that Lineplot doesn't use `tickformat` so it's not a simple story.

I thought I'd just comment so you could investigate further. I'll try to diagnose more tomorrow if I have time.

bobular:
Here's the old ticket, which should help for context
https://github.com/VEuPathDB/web-components/issues/185

moontrip:
@bobular thank you for your review 👍 I will check it out too

moontrip:
@bobular I have tested with lineplot component and its Viz following your example. I have also made a lineplot story for log scale to more accurately test under storybook env. I have concluded that actually an axis prop, dtick, which was implemented by you to have at least one tick, causes the issue you noticed. If it is commented out, the output is like the below, which is just fine.
I have committed two files lineplot by commenting out yaxis.dtick and lineplot story with log scale. You can test them at storybook. Improving dtick logic would help but it may not be simple task to satisfy all possible cases, especially including the axis range changes by users. What do you think?

lineplot-logscale1

bobular:
Hi @moontrip thank you for looking into this further. I don't have time to go into it until mid next week, but I did notice in your screenshot that that is NOT a log scale on the y-axis. The numbers can't be equally spaced like that!

moontrip:
@bobular I think that it is correct behavior. The interval of the range is quite small, less than 10, so it looks like equally spaced. For comparison purpose, I attached two screenshots with different ranges taken from storybook (lineplot story I added), such as [32,38] and [20,40]. The first one looks like equally spaced but if you increase the range, e.g., 20 to 40, then it looks like a log scale. 


lineplot-logscale2-1

lineplot-logscale2

bobular:
Thanks for the Lineplot storybook - I've been having problems with relative-deps on my slow laptop, so this is great.

Have you noticed that currently, the "nasty" 2 and 5 tick marks are back?  I say "nasty" because the ClinEpi folk didn't like them. I'm not a huge fan either, to be honest.

image

moontrip:
@bobular Thank you for your tests and comments. I think that 2,5 tick marks come from automatic dtick configuration from plotly.js. I think that we can remove that 2, 5 for such a range in log scale to be 10^-1, 1, 10, 100 when we set dtick to 1 when log scale is on?

moontrip:
@bobular I have revised the logic the way we discussed at Slack. So, mostly it works fine. Just one thing I noticed an edge case that with the condition to check dtick < 1, the y-axis tick label may have long decimal places like your latest screenshot. I tried to trim the decimal places in several ways, but from my understanding, the only way is to set tickformat attribute. However, this is conflicted with the use of power format, thus we may just live with it.

moontrip:
New logic is applied to other plots too.

moontrip:
@bobular Thank you for your tests and reviews 👍

PR #385, opened 2022-08-10
changing color for map continuous overlay

moontrip:
This will possibly address https://github.com/VEuPathDB/web-eda/issues/1200. The easiest solution is to change the order of colors defined in SequentialGradientColorscale: alternatively, one may adjust array index at mapViz but in this PR, the former is used.

The change of color order at legend means that the bar colors should also be changed. Here is an example screenshot to show.
map-continuous-color

(did not clean up the PR yet but will do it later)

### DIFF
--- a/src/types/plots/addOns.ts
+++ b/src/types/plots/addOns.ts
@@ -145,18 +145,32 @@ export const ColorPaletteDark: string[] = [
 
 /** Sequential gradient colorscale. Useful for coloring based on a continuous variable that is always positive, for example. */
 /** Using oslo from https://www.fabiocrameri.ch/colourmaps/, copied from https://github.com/empet/scientific-colorscales/blob/master/scicolorscales.py */
+//DKDK test
+// export const SequentialGradientColorscale: string[] = [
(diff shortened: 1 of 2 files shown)


src/types/plots/addOns.ts
```diff
@@ -145,18 +145,32 @@ export const ColorPaletteDark: string[] = [
 
 /** Sequential gradient colorscale. Useful for coloring based on a continuous variable that is always positive, for example. */
 /** Using oslo from https://www.fabiocrameri.ch/colourmaps/, copied from https://github.com/empet/scientific-colorscales/blob/master/scicolorscales.py */
+//DKDK test
```
bobular, line 148: OK to remove these commented lines now?

bobular (approved):
Looks good to me!

src/types/plots/addOns.ts
moontrip: Welcome back @bobular I hope you enjoyed your vacation! Thank you for your review and comment. I will comply with and then merge it (because it is quite simple to fix 😄)

PR #388, opened 2022-08-17
Use existing default string for x-axis label in ChartMarker popover plot

### LINKED ISSUE #386: Long axis title truncated on barplot in Maps
Long axis title was truncated on barplot in map visualization. 
It should read "181 Community repeated measures"

Screen Shot 2022-08-16 at 2 34 58 PM

jernestmyers:
Resolves #386 

The current implementation uses the `Total: <insert count>` string as a default value for the independent axis, like so:
```js
independentAxisLabel={
  props.independentAxisLabel ?? `Total: ${sumValuesString}`
}
```
I've removed the option to render the passed-in `props.independentAxisLabel` to only render the default string. The result is as shown.
image

### DIFF
--- a/src/map/ChartMarker.tsx
+++ b/src/map/ChartMarker.tsx
@@ -219,7 +219,6 @@ export default function ChartMarker(props: ChartMarkerProps) {
 
   // anim check duration exists or not
   let duration: number = props.duration ? props.duration : 300;
-  // let duration: number = (props.duration) ? 300 : 300
 
   const plotSize = 200;
   const marginSize = 5;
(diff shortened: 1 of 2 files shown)


moontrip (approved):
Well, it is quite easy to review 👍

src/map/ChartMarker.tsx
```diff
@@ -252,9 +251,7 @@ export default function ChartMarker(props: ChartMarkerProps) {
       displayLibraryControls={false}
       interactive={false}
       dependentAxisLabel=""
-      independentAxisLabel={
-        props.independentAxisLabel ?? `Total: ${sumValuesString}`
```
dmfalke, line 256: Should we also remove the `independentAxisLabel` prop from the `ChartMarkerProps` interface?

jernestmyers: I debated this and decided to leave it in for now with the thought there's likely future work to be done with the label. I'd prefer to find a way to keep the popup from disappearing such that it's possible to maybe hover over the axis and gain more information/context. But that comes at a much higher cost from what I can tell.

Then again, it's easy enough to add back the prop if/when the UI changes and it makes little sense for it to remain in the interim so I'll just go ahead and remove it.

dmfalke (approved):

PR #389, opened 2022-08-25
Barcharts small bar mouseover problem

bobular:
Just a draft at the moment, but we could merge, even if we don't find a solution.

I added a story to demonstrate.

I googled and tried a few things (including very large border widths) but nothing worked :sob:

### DIFF
--- a/src/stories/plots/Barplot.stories.tsx
+++ b/src/stories/plots/Barplot.stories.tsx
@@ -34,6 +34,25 @@ Basic.args = {
   title: 'Awesomeness of animals stratified by domestication',
 };
 
+export const TinyBar = Template.bind({});
+TinyBar.args = {
+  data: {
+    series: [
+      {
+        label: ['dogs', 'cats', 'monkeys'],
+        value: [10000, 10, 1],
+        name: 'Maybe',
+      },
+    ],
+  },
+  dependentAxisLabel: 'Awesomeness',
+  independentAxisLabel: 'Animal',
+  legendTitle: 'Domesticated',
+  opacity: 0.75,
+  interactive: true,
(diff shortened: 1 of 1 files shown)


moontrip (approved):
tested the story and confirmed the example that does not work for mouseover for small values

bobular:
> tested the story and confirmed the example that does not work for mouseover for small values

Thanks - I'll leave it here for a week or so and probably merge after next week's Wednesday meeting (and some discussion)

moontrip:
@bobular I think that if we adjust max range at dependent (y) axis range control, saying 100, then perhaps mouseover would work as small value bars are scaled up accordingly?

bobular:
This needs some UX input I think. @danicahelb @SheenaTomko @nkittur-uga

moontrip:
> > @bobular I think that if we adjust max range at dependent (y) axis range control, saying 100, then perhaps mouseover would work as small value bars are scaled up accordingly?
> 
> Yes, that would make the bar bigger and clickable. It's a workaround, just like switching on log scale.
> 
> Another possibility is to add a toggle switch to turn on/off displaying the values on the barchart. Both Barplot and Histogram have a `showValues` prop.
> 
> I just hacked this in browser dev tools (set `showValues: true`) and it looks quite good
> 
> image

@bobular I like the idea to show values but isn't it troubled with overlay variable? (stacked in histogram and grouped in barplot) Perhaps showValues: false for overlay variable? 😃

bobular:
It's a bit of a mess with an overlay, yes

image

danicahelb:
@bobular I don't think i like the toggle to display values. we will likely add a toggle to add counts in each population and all of these different numbers will get very confusing!

PR #390, opened 2022-08-25
bug fix and refining

moontrip:
While working on updating the long lasting react-leaflet v3, I realized that ChartMarker has a bug which causes issue with its story file. This did not cause an issue at Map viz in most cases. Also refined a bit for existing code.

I have checked this with Map viz and the changes worked just fine.

### DIFF
--- a/src/map/ChartMarker.tsx
+++ b/src/map/ChartMarker.tsx
@@ -221,7 +221,7 @@ export default function ChartMarker(props: ChartMarkerProps) {
   });
 
   // anim check duration exists or not
-  let duration: number = props.duration ? props.duration : 300;
+  const duration: number = props.duration ? props.duration : 300;
 
   const plotSize = 200;
   const marginSize = 5;
--- a/src/map/ChartMarker.tsx
+++ b/src/map/ChartMarker.tsx
@@ -140,7 +140,7 @@ export default function ChartMarker(props: ChartMarkerProps) {
 
   // drawing bars per marker
   fullStat.forEach(function (
-    el: { color: string | undefined; label: string; value: number },
+    el: { color: string; label: string; value: number },
     index: number
   ) {
     // for the case of y-axis range input: a global approach that take global max = icon height
@@ -170,7 +170,7 @@ export default function ChartMarker(props: ChartMarkerProps) {
       barHeight +
       ' fill=' +
       // rgb strings with spaces in them don't work in SVG?
-      el.color?.replace(/\s/g, '') +
+      el.color.replace(/\s/g, '') +
       ' />';
   });
 
--- a/src/map/ChartMarker.tsx
+++ b/src/map/ChartMarker.tsx
@@ -88,14 +88,17 @@ export default function ChartMarker(props: ChartMarkerProps) {
         });
 
   // determine min/max (one marker)
-  const minMaxPosRange: NumberRange = props.dependentAxisRange
-    ? props.dependentAxisRange
-    : {
-        min: props.dependentAxisLogScale
-          ? Math.min(...fullStat.map((o) => o.value).filter((a) => a > 0))
-          : 0,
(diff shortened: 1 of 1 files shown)


src/map/ChartMarker.tsx
```diff
@@ -136,40 +139,42 @@ export default function ChartMarker(props: ChartMarkerProps) {
...
-      ? el.value <= 0
-        ? 0
-        : // if dependentAxisRange != null, plot with global max
-          (Math.log10(el.value / minMaxPosRange.min) /
-            Math.log10(minMaxPosRange.max / minMaxPosRange.min)) *
-          (size - 2 * marginY)
-      : ((el.value - minMaxPosRange.min) /
-          (minMaxPosRange.max - minMaxPosRange.min)) *
-        (size - 2 * marginY); // bar height: used 2*marginY to have margins at both top and bottom
+  fullStat.map(
```
bobular, line 142: Is there a deep philosophical reason to change from `forEach` to `map`, or just a style preference?

src/map/ChartMarker.tsx
```diff
@@ -88,14 +88,17 @@ export default function ChartMarker(props: ChartMarkerProps) {
...
-    ? props.dependentAxisRange
-    : {
-        min: props.dependentAxisLogScale
-          ? Math.min(...fullStat.map((o) => o.value).filter((a) => a > 0))
-          : 0,
-        max: Math.max(...fullStat.map((o) => o.value).filter((a) => a > 0)),
-      };
+  const minMaxPosRange: NumberRange =
+    props.dependentAxisRange != null &&
+    props.dependentAxisRange.min != null &&
```
bobular, line 93: You didn't mention what the bug was, but presumably checking for `min` and `max` `!= null` was the main fix?

bobular (approved):
This seems OK. All the ChartMarker stories work for me. Were they broken before?

How will we deploy and QA this?  Maybe we need a bug ticket anyway?

src/map/ChartMarker.tsx
moontrip: @bobular Thank you for reviews and tests!
For this item, yeah just a style preference and no specific reason - we hardly used forEach at our codes so just unified it 😄

src/map/ChartMarker.tsx
moontrip: yes so it was not sufficient to check in the previous code when no dependentAxisRange.min and dependentAxisRange.max were not set. It is because we have changed the logic with you, using dependentAxisRange.min & .max when computing barHeight: if they are undefined, no vertical bars are not shown in the chart marker

src/map/ChartMarker.tsx
dmfalke: I appreciate wanting to unify the coding style, but I don't think it's a good idea to use `.map` if you are modifying variables outside the scope of the callback function. It's meant to be used to construct a new array, and so can confuse someone else reading the code. Using `.forEach` will be a better signal to the reader for what the callback function is doing.

moontrip: @dmfalke thank you for your suggestion. 👍 I got your point for this case thus I changed it back to use .forEach (easy one 😄)

src/map/ChartMarker.tsx
```diff
@@ -167,7 +170,7 @@ export default function ChartMarker(props: ChartMarkerProps) {
       barHeight +
       ' fill=' +
       // rgb strings with spaces in them don't work in SVG?
-      el.color.replace(/\s/g, '') +
+      el.color?.replace(/\s/g, '') +
```
bobular, line 173: So this could end up being `fill=` (with nothing after it, if `el.color` is undefined) - is that valid SVG?

moontrip: as I said at Slack, I have removed undefined in type and committed 👍

moontrip:
This PR is closed as https://github.com/VEuPathDB/web-components/pull/445 is merged

PR #395, opened 2022-09-11
Full screen map: minimap

bobular:
Still work in progress


dmfalke (approved):

src/map/MapVEuMap.tsx
```diff
@@ -297,7 +325,7 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
         recenterMarkers={recenterMarkers}
       />
 
-      {showMouseToolbar && (
+      {showMouseToolbar && mouseMode && onMouseModeChange && (
```
moontrip, line 328: @bobular I may be wrong but I don't think these additional conditions are necessary - you can just control showMouseToolbar (boolean). See my PR for FSM, https://github.com/VEuPathDB/web-eda/pull/1357

bobular: Hi @moontrip - you are probably right, but my IDE was giving me TS warnings about `mouseMode` and `onMouseModeChange` being undefined in the props below :-(

moontrip: Hi @bobular 👋 Aha I see. That makes sense as the minimap does not have mouseMode. Thank you for your kind explanation 👍

PR #397, opened 2022-09-14
RadioButtonGroup - new onSelectedOptionDisabled prop

bobular:
Needed for work on FSM overlay variable constraints (Bob's PR pending)

New functionality and new story to demonstrate the situation where you have disabled a radio button input after it has already been selected. You can provide the `onSelectedOptionDisabled` callback to perform an action, such as setting the radio button state to 'default' as in the story. You could also show a snackbar! Mmm, snack bars...

### DIFF
--- a/src/components/widgets/RadioButtonGroup.tsx
+++ b/src/components/widgets/RadioButtonGroup.tsx
@@ -1,5 +1,5 @@
 // widget for radio button group
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useEffect } from 'react';
 import Radio from '@material-ui/core/Radio';
 import RadioGroup from '@material-ui/core/RadioGroup';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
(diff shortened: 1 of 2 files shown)


dmfalke (approved):

src/components/widgets/RadioButtonGroup.tsx
```diff
@@ -19,6 +19,8 @@ export type RadioButtonGroupProps = {
   selectedOption: string;
   /** Action to take when an option is selected by the user. */
   onOptionSelected: (option: string) => void;
+  /** Action to take if the currently selected option has been disabled. */
+  onSelectedOptionDisabled?: (option: string) => void;
```
jernestmyers, line 23: Small nit, but I wonder if we should be leveraging a `not-allowed` cursor when radio buttons are disabled? Not a showstopper.

jernestmyers (approved):
Looks good, and yes: there's an [open issue in `CoreUI` for radio buttons](https://github.com/VEuPathDB/CoreUI/issues/66).

src/components/widgets/RadioButtonGroup.tsx
bobular: Thanks - that's a nice thing to add (and I added it).

bobular:
I'm not merging this right now.  Working on a different approach EDA-side.  I will switch this to Draft mode

PR #402, opened 2022-10-25
Unify plot controls styling and verbiage

jernestmyers:
Works towards https://github.com/VEuPathDB/EdaNewIssues/issues/456

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "@types/react-plotly.js": "^2.2.4",
     "@types/stats-lite": "^2.2.0",
     "@veupathdb/browserslist-config": "^1.0.0",
-    "@veupathdb/coreui": "^0.18.0",
+    "@veupathdb/coreui": "^0.18.11",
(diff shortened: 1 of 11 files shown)


moontrip (commented):
@jernestmyers overall looks great! 👍 
I noticed one minor issue for a Lineplot case. If xAxisVariable is categorical, all X-axis controls are grayed out. However, the texts for axis range control (Full, Auto-zoom, and Custom) still remain blackish. Here are screenshots to compare between QA site and local VM:

Didn't check code carefully, but perhaps you did set those text to be black even for disabled case? (Radio button group?)

- QA site
lineplot-qa1

- Local VM
lineplot-t1

jernestmyers:
Very good catch! Just pushed a fix.

moontrip:
@jernestmyers I just confirmed that it worked out 👍 Will approve both PRs

moontrip (approved):

PR #404, opened 2022-10-27
Avoid extra bin labels

bobular:
Resolves https://github.com/VEuPathDB/web-eda/issues/872

I decided to keep this on the web-components side, and only make changes to the axis configuration when there are a tiny number of bins (<= 4).  If we need to adjust the threshold (or the approach) after testing in the wild, that's fine. I just wanted to tread as lightly as possible to avoid unintended consequences.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.19.5",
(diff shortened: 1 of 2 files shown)


chowington (approved):
Looks good, @bobular! 👍

PR #405, opened 2022-10-27
Fix red warning for bin width at limit of range

bobular:
Resolves https://github.com/VEuPathDB/web-eda/issues/872

Solves the issue where the numeric bin width slider input is outlined red when at an allowed lower or upper limit.
Adjusting the bin width to the min or max was only possible with the slider, not the number input.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.19.6",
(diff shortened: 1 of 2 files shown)


moontrip (approved):
@bobular I didn't know if that was the case 👍 I couldn't see red box anymore

PR #406, opened 2022-11-03
handling html tags at y-axis label tooltip

moontrip:
This PR is related to https://github.com/VEuPathDB/web-eda/issues/1433

I figured it out that dependent axis label can take html tags like `<b>, <i>, <br>`, however, its tooltip, which is actually SVG text attribute, still shows such tags as string. Thus, it is necessary to remove tags for tooltip string.

I also made a story to test this, as shown in the following screenshot. Those tags are not shown in the tooltip.

lineplot-yAxis-label1

### DIFF
--- a/src/plots/PlotlyPlot.tsx
+++ b/src/plots/PlotlyPlot.tsx
@@ -27,6 +27,7 @@ import { ToImgopts, toImage } from 'plotly.js';
 import { uniqueId } from 'lodash';
 import { makeSharedPromise } from '../utils/promise-utils';
 import NoDataOverlay from '../components/NoDataOverlay';
+import { removeHtmlTags } from '../utils/removeHtmlTags';
 
 export interface PlotProps<T> extends ColorPaletteAddon {
   /** plot data - following web-components' API, not Plotly's */
@@ -223,7 +224,7 @@ function PlotlyPlot<T>(
         .selectAll('g.traces')
         .append('svg:title')
         .text((d: any) => {
-          return storedLegendList[d[0].trace.index];
+          return removeHtmlTags(storedLegendList[d[0].trace.index]);
         });
 
       // legend title tooltip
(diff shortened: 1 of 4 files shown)


src/plots/PlotlyPlot.tsx
```diff
@@ -293,7 +293,13 @@ function PlotlyPlot<T>(
           // need this attribute for tooltip of dependent axis title!
           .attr('pointer-events', 'all')
           .append('svg:title')
-          .text(originalDependentAxisTitle as string);
+          // remove html tags
+          .text(
+            (originalDependentAxisTitle as string).replace(
+              /[<b></b><i></i><br>]/g,
```
bobular, line 299: This is removing the letters 'i', 'b' and 'r' from the tooltip as well! (the regexp square brackets form a character class).

I'd suggest using a html stripping library (I don't have a recommendation, sorry!) or something like this

```
  regexp = new RegExp("(?:</?b>|</?i>|<br\\s*/?>)", "gi")
```

The `?:` is optional, but good practice to not "capture" (e.g. into $1, $2) the match unless you're planning to do something with it.

I added a 'i' modifier which should make it case-insensitive (in Perl it does anyway).

I used the `new RegExp` constructor so we don't need to escape the forward slashes IN the regexp, which makes it a bit more readable.

I extended the br pattern to match `<br />` too.

bobular (changes requested):
See my suggestion.

Also do we want to document that the `dependentAxisTitle` can contain certain HTML tags?

src/plots/PlotlyPlot.tsx
bobular: I'd also suggest making a function to strip simple HTML tags from all tooltip texts, rather than just for this axis label.

moontrip: @bobular Oops! yes you are right. Thank you for your suggestions! 👍 I will check them out.

moontrip: @bobular I have updated the regexp. Your regexp worked fine with the b, i, br tags so I used it. In addition, I also added another regexp, which is commented out though, for handling more general case provided that those tags are in the format of `<tags>`.

bobular (approved):

PR #409, opened 2022-11-11
change the order of Mosaic color

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1435

Initially, more complicated approach was used to address this issue by changing PlotlyPlot component, however, it was found that there was much simpler way. 

Here is a screenshot using the same example case in the ticket:

mosaic-color1


chowington (approved):
Nice and simple! 👍

PR #418, opened 2022-11-28
Mosaic elbow labels

chowington:
Works toward https://github.com/VEuPathDB/web-eda/issues/617 and https://github.com/VEuPathDB/web-components/issues/427

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.19.12",
+  "version": "0.19.13",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
@@ -8,6 +8,8 @@
     "url": "https://github.com/veupathdb/web-components.git"
   },
   "dependencies": {
+    "@typescript-eslint/eslint-plugin": "^5.46.0",
+    "@typescript-eslint/parser": "^5.46.0",
     "@visx/gradient": "^1.0.0",
     "@visx/group": "^1.0.0",
     "@visx/hierarchy": "^1.0.0",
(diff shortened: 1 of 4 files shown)


chowington:
Realized my very dumb mistake after the meeting.....ready for review now!

moontrip (commented):
@chowington this looks good. I have found two or three potential issues as follows. Note that I used GEMS1 Case Control data for these tests.

a) I noticed that you introduced scroll for Y-axis legend, but it seems like it has an interference with the first(?) data: X-axis baseline shown in the bottom of the plot at the leftmost has more horizontal line than plot itself (or it may be due to other reason)
GEMS1-elbow1

b) The Y-axis legend scroll seems not to work at facet plot. 
GEMS1-elbow-facet1

c) Possibly another one, which may not be a bug, is that the rightmost X-axis baseline indicates data and three labels but seems like no data exist even at full screen modal: it may be an intended behavior or really small data values, though.
GEMS1-elbow-facet-modal1

chowington:
Thanks DK! I didn't intentionally add any scrollbar function; that appears if the legend is taller than the Y axis. I did, however, fix some logic that was making the gap between the legend items too big. There will be some cases where, even with no gap, the legend will be too large, and in those cases the scrollbar should appear.

Plotly apparently shrinks the first column to make room for the scrollbar, and I don't see an easy way to change that. Anyway, it doesn't seem to shift any of the other columns. Also, I don't know why Plotly didn't show a scrollbar for those faceted plots---maybe it didn't quite think the legend was too tall yet. Anyway, I don't think there's an easy way to fix that either; it should happen less thanks to my latest commit though!

chowington:
Oh, and yes, the labels bunched up at the end I think is expected behavior, since they have very few values.

moontrip (approved):
Thank you for your detailed explanations, @chowington 👍 Yes I didn't notice the issue I had. We will see how the testers will say 😄

PR #422, opened 2022-12-05
FSM marker-related work

### LINKED ISSUE #416: Full screen map: refactor donut and chart markers to allow them to also be rendered outside Leaflet
In a donut marker story, I have tried accessing `markers[0].props.icon` which ought to contain a `Leaflet.divIcon` that should have a `createIcon` method. But the `icon` prop is `null` when I try to access it. So I don't recommend trying that route any further.

The cleaner solution anyway would be to factor out the `svgHTML` generation in DonutMarker.tsx and ChartMarker.tsx into helper functions (one takes `DonutMarkerProps`, the other takes `ChartMarkerProps` as arguments - excluding the `onClick` prop, which seems to be unused anyway [should we remove this?]).  

Then create two new Leaflet-free components: DonutMarkerIcon and ChartMarkerIcon that simply render a div with the SVG icon in it (using the new helper functions of course).  The new components take the same props as the regular markers (especially if we get rid of the `onClick`).

### LINKED ISSUE #421: Full screen map - cumulative mode for DonutMarker
Add a new prop `cumulative?: boolean` to `DonutMarker`

When true, when given label/value pairs  `filtered=85, total=100`, would display a 85% pink and 15% grey donut marker. Mouse-overs would show the correct numbers: 85, and 100.

Do we assume the values are sorted already? Probably yes. This is because when provided with `filtered=100, total=100` we want the `filtered` "segment" of the donut to come first/be coloured and leave the `total` segment invisible/size zero.

So in summary, when `cumulative == true`, assume incoming values are in order, the final value is the total, and any values before that are fractions of that.

bobular:
- [x] Resolves #421 
- [x] Resolves #416 

Question for reviewers: is the mouse pointer behaviour on the standalone markers what we want? I guess it is just the default behaviour because I don't think we've done anything special? Update: now I've done the chart markers too, the mouse pointer behaviour is completely different for that! It would be good if there was no special mouse pointer behaviour.  Suggestions welcome!

Also are we happy with the dangerous setting of inner HTML?

I think the UX people are not going to be happy with the cumulative donut markers, but we will cross that bridge when we come to it...

### DIFF
--- a/src/map/DonutMarker.tsx
+++ b/src/map/DonutMarker.tsx
@@ -11,10 +11,10 @@ import {
   MarkerScaleAddon,
   MarkerScaleDefault,
   PiePlotDatum,
+  ContainerStylesAddon,
 } from '../types/plots';
 
 import { last } from 'lodash';
-import { ContainerStylesAddon } from '../types/plots';
 
 //DKDK ts definition for HistogramMarkerSVGProps: need some adjustment but for now, just use Donut marker one
 export interface DonutMarkerProps
--- a/src/map/DonutMarker.tsx
+++ b/src/map/DonutMarker.tsx
@@ -14,6 +14,8 @@ import {
   PiePlotDatum,
 } from '../types/plots';
 
+import { last } from 'lodash';
+
 //DKDK ts definition for HistogramMarkerSVGProps: need some adjustment but for now, just use Donut marker one
 export interface DonutMarkerProps
   extends BoundsDriftMarkerProps,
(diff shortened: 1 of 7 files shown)


src/map/DonutMarker.tsx
```diff
@@ -256,3 +175,133 @@ export default function DonutMarker(props: DonutMarkerProps) {
...
+  return (
+    <div
+      style={{
+        fontSize: '12px',
+        lineHeight: 1.5,
+        width: size,
+        height: size,
+        ...props.containerStyles,
+      }}
+      dangerouslySetInnerHTML={{ __html: html }}
```
bobular, line 205: Any problem with this :point_up: @dmfalke ?

moontrip (commented):
@bobular Looks good to me 👍 By the way, one question is what the cumulative means in this case as I am confused. Suppose that we have three data values, 10, 85, and 100 (from one of the data in the story). If cumulative is true, I thought then the values for drawing the arcs would have been 10, 95 (10+95), and 195 (10+85+100), but based on the code, it is 10, 75, and 25 and the fullPieValue to draw the proportional arc is 100, not 195. Perhaps the meaning of the cumulative is somewhat different from a typical meaning?

p.s. Well, you make me work for additional merge (and resolving conflicts) for my react-leaflet-v3 😄

bobular:
> Suppose that we have three data values, 10, 85, and 100 (from one of the data in the story). If cumulative is true, I thought then the values for drawing the arcs would have been 10, 95 (10+95), and 195 (10+85+100), but based on the code, it is 10, 75, and 25 and the fullPieValue to draw the proportional arc is 100, not 195. Perhaps the meaning of the cumulative is somewhat different from a typical meaning?

Hi @moontrip - the idea was that the values are **already** cumulative in nature. It's complicated because a regular pie chart is kind of cumulative by default: given values 40 and 60, the pie slice boundaries are drawn at 40 and 100 (or 60 and 100 if drawn in a different order).

I knew when I was commenting on the `cumulative` prop that I wasn't explaining very well. I will try to improve the comments.

src/map/DonutMarker.tsx
bobular: Yes, that was my feeling.

moontrip:
@bobular Thank you for your detailed explanation which is very helpful for me to understand it.

moontrip (approved):

PR #423, opened 2022-12-05
Mosaic long axis title

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1377. Since there is another mosaic work by Connor, https://github.com/VEuPathDB/web-components/pull/418, this is branched out from [mosaic-elbow-labels](https://github.com/VEuPathDB/web-components/tree/mosaic-elbow-labels) to minimize conflicts and accommodate new layouts. Thus, this PR will be finalized when https://github.com/VEuPathDB/web-components/pull/418 is done.

Now both X and Y axes have ellipsis. For this, the max length of X-axis string is set to 80. Also, a story is made for testing. Here is a screenshot.

mosaic-axis-ellipsis

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.19.14",
+  "version": "0.19.15",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
--- a/src/plots/PlotlyPlot.tsx
+++ b/src/plots/PlotlyPlot.tsx
@@ -132,6 +132,9 @@ function PlotlyPlot<T>(
     [displayLibraryControls, interactive]
   );
 
+  const xAxisTitle = plotlyProps?.layout?.xaxis?.title;
+  const yAxisTitle = plotlyProps?.layout?.yaxis?.title;
+
   const finalLayout = useMemo(
     (): PlotParams['layout'] & LayoutLegendTitle => ({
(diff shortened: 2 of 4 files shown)


moontrip:
@chowington I was waiting for your elbow work to be done for this PR :) Since it is finished, I updated this PR and changed it to be ready for review

chowington:
Hi DK, thanks for this work! While testing I found a case where the title still overflows:

image

This is on the WASH Benefits Bangladesh Cluster Randomized Trial study. Can you look into this case? I wonder whether it would be worth it to have the title length dependent on the plot container size? That gets hairy sometimes though...

chowington:
Also, I'd suggest trying to DRY up this code bit. For instance, it could be helpful to factor out the ellipsis logic into a function, as well as to assign `plotlyProps?.layout?.xaxis?.title` to a const so you don't have to repeat it every time.

chowington (changes requested):
Changes requested above ☝️

moontrip:
@chowington Thank you for your reviews and tests! 👍 I have addressed them in the new commit: reduced the max length of x-axis title (attached a screenshot with the same example) and make that ellipsis of the axis title be a function, axisTitleEllipsis().

mosaic-long-axis-title1

chowington (approved):
Looks good, thanks DK!

moontrip:
> Looks good, thanks DK!

@chowington Thank you for your review and tests! 👍

PR #431, opened 2022-12-15
add a lineplot prop to enable connecting points

moontrip:
This is to define a type for a new lineplot prop, which will be used for https://github.com/VEuPathDB/web-eda/issues/1519.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.20.0",
+  "version": "0.20.1",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
--- a/src/types/plots/lineplot.ts
+++ b/src/types/plots/lineplot.ts
@@ -20,6 +20,8 @@ export type LinePlotDataSeries = Override<
   extraTooltipText?: string[];
   /* Strings (such as '[10,15)') to describe how the x-axis variable was binned for tooltip info only */
   binLabel?: string[];
+  /* for connecting points regardless of missing data */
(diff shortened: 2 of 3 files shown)


dmfalke:
Would another way to address this be to remove the "missing data" points from the data series? 

Or, alternatively, is there ever a reason to not have these lines? If not, then should we always proceed to connect points, which would obviate the need for this prop?

dmfalke:
I see in your description of https://github.com/VEuPathDB/web-eda/pull/1531 that removing the missing data is too complex.

My second question stands: can we just assume we always want points connected?

moontrip:
@dmfalke Although I said it would be too complex, I may need to consider it depending on another ticket, https://github.com/VEuPathDB/web-eda/issues/1082. I didn't look into the requirements carefully yet though.

For your second question, I think yes? 😄 That's the reason why I did set it to be true at this corresponding web-eda PR: https://github.com/VEuPathDB/web-eda/pull/1531. The plotly prop should be set at data (processedData) level.

chowington:
Can't tell whether @Dave's question (`"can we just assume we always want points connected?"`) is resolved or not--- @danicahelb or @d-callan, any thoughts?

danicahelb:
@dmfalke @moontrip @chowington regarding 

> can we just assume we always want points connected? 

No. we have agreed with @d-callan that if the data series contains 0/0 there will be a break in the line. Perhaps we should wait until mockups for https://github.com/VEuPathDB/web-eda/issues/1082 are finalized and the new UI is implemented?

dmfalke (approved):

PR #433, opened 2022-12-19
change prop name

moontrip:
Following @bobular's suggestion on https://github.com/VEuPathDB/web-eda/pull/1494#issuecomment-1357876861, the prop name is changed to markerBodyOpacity.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.19.13",
+  "version": "0.19.14",
   "license": "Apache-2.0",
   "description": "React Components for VEuPathDB Websites",
   "repository": {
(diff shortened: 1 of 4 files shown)


moontrip:
@jernestmyers Thank you for finding a bug. 👍 I just committed a fix.
So the bug will not happen if the corresponding web-eda work (https://github.com/VEuPathDB/web-eda/pull/1494) will be merged. But since a marker opacity work at web-components was already merged, a logic to compute marker.line.width causes an issue without the web-eda PR. I have added to identify such an edge case so that the marker opacity works without the web-eda PR.

jernestmyers (approved):
This looks to resolve the missing markers in my analyses' scatterplots.
image

moontrip:
> This looks to resolve the missing markers in my analyses' scatterplots. image

@jernestmyers Thank you for your confirmation! 🥳

PR #434, opened 2022-12-23
Colormap stories

asizemore:
Adds a few stories to showcase the different colormaps used in the project. Currently in draft form, as we plan to discuss colormaps in the next data viz meeting. Will make adjustments after that meeting.

Anything else that would be helpful to see in these stories?

Note
- the Template is basically repeated in each file. I liked the organization of the colormaps into separate files more than i hated this repetition. 
- I removed one of the original scatterplot stories. It seemed not worth maintaining because it requires us to keep some older code in `processInputData`.
- On that note, i broke the versions of scatterplot stories that use a ribbon around a line (ex. smoothed mean CI). And by broke i mean the ribbon is opaque instead of transparent. I did not try to fix it because we aren't using these colors anymore and so these stories are not so useful anymore as they are. They could be made useful, but that feels like a different gh issue :) 
- added a slider that controls the number of points in the scatterplot
- I did not link the opacity slider and legend in this PR. Leaving for another gh issue.
- I used the actual opacity slider instead of a story argument since our slider also handles the change in marker


src/stories/colormaps/Categorical.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+    'markers',
+    'number',
+    'number',
+    false,
+    ColorPaletteDefault
+  );
+
+  // Reduce the input data to only approximately nPoints points. There will be multiple series so we have to
+  // slice each series at nPoints/nSeries.
+  for (
```
bobular, line 53: It's a pure style/pedantic suggestion, but we could avoid the `for` loop and modification of the contents of `datasetProcessCategorical` with something like this (not compiled or tested)

```
const trimmedDatasetCategorical = {
   ...datasetProcessCategorical,
   series: datasetProcessCategorical.series.map(({x, y, ...rest}) => ({x: x.slice(), y: y.slice(), ...rest})),
};
```

bobular (approved):
Still in draft but thought I'd approve anyway!

d-callan (approved):

asizemore:
finally all ready! @d-callan and @bobular i know you already approved but just did a re-request as a nudge in case you wanted a second look after the recent changes

Issue #432, opened 2022-12-19
Reimplement map marker SVG in JSX


bobular:
Donut and Chart markers SVG can be made with regular JSX.

Should obviate the need for the `dangerouslySetInnerHTML` we are using for standalone markers.

PR #444, opened 2023-01-30
fix a chartmarker bug

moontrip:
This will fix the bug I mentioned at https://github.com/VEuPathDB/web-eda/issues/1586

So, the issue was that some data/marker do not have color info, e.g., when data.value = 0 then no data.color exists, but svg does not like the empty value of `fill` attribute. Thus, I have added a dummy color (white) for no data.color. This manual assignment won't matter because if data.value = 0, then height of the data becomes 0 so its color is actually ignored anyway.

Here are screenshot after this fix: as compared to the problematic screenshots at the https://github.com/VEuPathDB/web-eda/issues/1586.

a) SCORE S. mansonni CRT study
map-bug-no-data-color-fixed

b) SCORE Mozambique (Full screen map)
mozambique map error fixed

### DIFF
--- a/src/map/ChartMarker.tsx
+++ b/src/map/ChartMarker.tsx
@@ -269,8 +269,8 @@ function chartMarkerSVGIcon(
       ' height=' +
       barHeight +
       ' fill=' +
-      // rgb strings with spaces in them don't work in SVG?
-      (el.color ?? '').replace(/\s/g, '') +
+      // empty string does not work: filled with white rgb
+      (el.color ?? 'rgb(255,255,255)').replace(/\s/g, '') +
       ' />';
   });
 
--- a/src/map/ChartMarker.tsx
+++ b/src/map/ChartMarker.tsx
@@ -270,7 +270,7 @@ function chartMarkerSVGIcon(
       barHeight +
       ' fill=' +
       // empty string does not work: filled with white rgb
-      (el.color ?? 'rgb(255,255,255)').replace(/\s/g, '') +
+      (el.color ?? 'rgb(192,192,192))').replace(/\s/g, '') +
       ' />';
   });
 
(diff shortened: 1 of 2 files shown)


src/map/ChartMarker.tsx
```diff
@@ -269,8 +269,8 @@ function chartMarkerSVGIcon(
       ' height=' +
       barHeight +
       ' fill=' +
-      // rgb strings with spaces in them don't work in SVG?
-      (el.color ?? '').replace(/\s/g, '') +
+      // empty string does not work: filled with white rgb
+      (el.color ?? 'rgb(255,255,255)').replace(/\s/g, '') +
```
bobular, line 273: I think the best thing to do here is use `'silver'` as the default. I don't particularly like it, but it's consistent with DonutMarker.  The problem is that the storybook story has a value of 0. If it was non-zero (which I think makes a better story), we would want the bar to be visible.  Otherwise we should make the `color` prop mandatory!

bobular (changes requested):

src/map/ChartMarker.tsx
bobular: Thanks for spotting the nasty bug!
The SVG syntax error kills the rest of the marker.

#432 will sort this, but sadly it's not a "must have" for the next release or two

moontrip: @bobular Thank you for your comments. It does not matter we use silver, white, etc. Although I didn't check how to create data/marker at useMapMarkers hook at web-eda, a general pattern concerning the bug was that data.color is undefined when data.value = 0. The story's data actually came from one of real data in the first screenshot case.

Anyway, I will change it to silver as it does not matter much to resolve this bug.

moontrip:
@bobular Just for a reminder, I have addressed your feedback before so it would be great if you can check this out again 👍

moontrip:
This PR is closed as https://github.com/VEuPathDB/web-components/pull/445 is merged: this bug was also applied at the PR

PR #447, opened 2023-02-02
add condition for dtick in case of no data

moontrip:
This PR is related to https://github.com/VEuPathDB/web-eda/issues/1503. It is only for a better display of log scale without data, so for a functionality wise, it does not matter much.


bobular (approved):
Good catch!

PR #449, opened 2023-02-07
no gray for legend when unchecked

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1602. Tested extensively with web-eda Vizs and all seem to work fine. 

Here are example screenshots with a Scatterplot Viz.

- using qa-restricted (unchecked legend item turns to be gray)
scatter legend qa1

- using dev with current PR (unchecked legend item remains the same color, not being gray)
scatter legend dev1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.19.18",
+  "version": "0.19.19",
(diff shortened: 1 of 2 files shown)


chowington (approved):
Looks good!

PR #450, opened 2023-02-09
implementing reverse slider feature

moontrip:
This is related to https://github.com/VEuPathDB/EdaNewIssues/issues/541. 

I missed the latest comment on the ticket so I implemented a new feature, reverse slider, which can be used if necessary. To utilize a reverse slider, two settings should be changed. 

a) isReverseSlider: true (default is false)
b) change trackGradientStart and trackGradientEnd to reflect appropriate colors

Added a story at Slider.stories.tsx to take a look at how it works.

A screenshot is attached. 
reversed slider01

### DIFF
--- a/src/components/widgets/Slider.tsx
+++ b/src/components/widgets/Slider.tsx
@@ -1,5 +1,9 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
-import { makeStyles } from '@material-ui/core/styles';
+import {
+  makeStyles,
+  createMuiTheme,
+  ThemeProvider,
+} from '@material-ui/core/styles';
 import Slider from '@material-ui/core/Slider';
 import Typography from '@material-ui/core/Typography';
 
@@ -53,6 +57,8 @@ export type SliderWidgetProps = {
   showLimits?: boolean;
   /** Disable the slider. Default is false */
   disabled?: boolean;
+  /** is reversed slider? e.g., from max to min */
+  isReverseSlider?: boolean;
 };
 
 /** A customizable slider widget.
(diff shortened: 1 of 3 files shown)


moontrip:
@jernestmyers if you have time, it would be great if you can review this PR 👍

jernestmyers (commented):
I'm struggling to think of a use-case for this reversed state. To that end, I'd rather defer to other folks to see if they can imagine a use-case for these changes. If not, then I'd suggest we scrap it. If so, we can approve and move forward.

moontrip:
> I'm struggling to think of a use-case for this reversed state. To that end, I'd rather defer to other folks to see if they can imagine a use-case for these changes. If not, then I'd suggest we scrap it. If so, we can approve and move forward.

@jernestmyers Thank you for your review and suggestion! 👍 Actually the original ticket strongly requested to have the reverse slider for marker opacity: see this (https://github.com/VEuPathDB/web-eda/issues/1487#issuecomment-1380784054) and https://github.com/VEuPathDB/web-eda/issues/1605. Things were changed during I suffered from sickness in the end of Jan. that Danica just wanted to remain the slider as is. I found her comment late, after implementing the feature of reverse slider.

Honestly I am not quite sure whether they want to go back to the reverse slider or not. If you think this is overkill, then perhaps we can just switch this PR to draft one. The corresponding web-eda PR (https://github.com/VEuPathDB/web-eda/pull/1606) does not necessarily need this PR if a normal slider is used.

moontrip:
> Yeah, I think we should ask those involved in the request if they can envision us wanting this option in the future. The code itself looks great, I just want to be sure we have a need for this behavior first!

Sounds great @jernestmyers 👍 Perhaps we can leave this as is for the time being as this is not essential part of b62 anyway 😄  I will ask them about it at the original ticket's comment

Issue #454, opened 2023-01-27
legend updates needed for divergent colormaps for continuous overlay variables 


danicahelb:
From @bobular 

In GEMS1, BMI-for-age z-score ranges from -15 to +273, but when i use this term as an overlay, the legend indicates it ranges from -30 to +30. (I thought it could have something to do with not having data for the participant whose BMI-for-age z-score was 273 but checked and this is not the case)

I've had a quick look. The code is creating a symmetrical gradient around zero and is using the curated `displayRangeMin` and `displayRangeMax` in preference to the data-derived `rangeMin` and `rangeMax`.  (This preference has been standard practice until now, I believe.) So this is why we get -30 to 30 (because it creates a symmetric range using `max(abs())`

![image](https://user-images.githubusercontent.com/308639/205291017-65e6ed1d-da1f-4982-8f9a-0b894a3434dd.png)

<img width="1111" alt="image" src="https://user-images.githubusercontent.com/25450900/215169254-c8e051e7-a354-4799-9ba3-52c3eb7d18c1.png">

<img width="929" alt="image" src="https://user-images.githubusercontent.com/25450900/215169336-599a47cc-cdd1-41ba-919f-3722fd6dc7b5.png">

moontrip:
Hi @danicahelb May I ask two questions to clarify?

First, for the case of Bob's screenshot above, displayMinMax = (-30, 30) and rangeMinMax = (-15.3, 273.91), then should the new range be (-30, 273.91)? Second, do we need to consider subsetted (flitered) data? For example, assuming that the range of the displayMinMax and rangeMinMax are the same with the first question and somehow subsetted data has MinMax = (10, 20), then do we need to respect the combination of the displayMinMax and rangeMinMax (e.g., -30, 273.91) or just use data-based MinMax (10, 20) for the colormap range?

> In 1/25/2023 data viz group, we discussed how to set the default range for gradient overlays and how does it map to hues in the gradient
> 
> decision is that the gradient colormap should be consistent w what we do for the axes:
> 
> “Axes representing user-selected continuous variables will scale according to study metadata (meaning use the annotated display range from the owl file when available, expand to the full data set range if we see any values in the response which exceed that)”
> 
> Unlike what we do for the axes, if the resulting range does not include 0 we should NOT expand the range to do so. currently we do force the gradient map to run through 0 even when the data doesn't and this makes it hard to distinguish the colors.

bobular:
[...]

PR #464, opened 2023-03-21
set decimal points for colormap ticks

moontrip:
This is related to https://github.com/VEuPathDB/web-components/issues/454. 

While I was working on the colormap range, I noticed that the decimal points in the colormap's tick labels were often too long, so I have set it up to be 2 decimal points. Following is an example based on storybook.

a) without setting decimal points
decimal-points-no

b) two digit decimal points
decimal-points-2

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/components",
-  "version": "0.20.4",
+  "version": "0.20.5",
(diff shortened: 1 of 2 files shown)


bobular (changes requested):
Could we do some testing with `toPrecision` please?

